PR #7, opened 2020-07-31
Remove deprecated flags

vorot93:
This PR removes deprecated flags, along with vestiges of geth compatibility mode.


parity/rpc_apis.rs
```diff
@@ -298,9 +267,6 @@ impl FullDependencies {
 						&self.miner,
 						&self.external_miner,
 						EthClientOptions {
-							pending_nonce_from_queue: self.geth_compatibility,
```
adria0, line 301: I think that in this step is better to keep this `--geth` flag active, removing this flag means that maybe some users will have to change their code to call the RPC if they are running in this mode for whatever reason (for instance because they have two active/active nodes, one with geth and another with OE, and wants to use the same code to call/parse API calls)

So, is ensuring that there's no breaking changes in user space. Nonetheless I'm in that this flag should be removed in a near future.

vorot93: The main reason I'm in for the removal is that we don't know if these quirks are still valid, and we don't have the capacity to actively monitor geth's RPC behaviour. Hence it's only honest to users to be upfront about not supporting these quirks.

adria0: What about renaming the flag to `--geth-deprecated` and generate a big warning banner what this will be deprecated in next version?

parity/cli/tests/config.full.toml
```diff
@@ -34,10 +33,6 @@ sstore_url = "http://localhost:8082"
 sstore_threshold = 0
 
 [ui]
-force = false
-disable = false
-port = 8180
-interface = "127.0.0.1"
 path = "$HOME/.parity/signer"
```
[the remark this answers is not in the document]
vorot93: Yes, but that also pulls in the removal of Signer interface which will happen in a subsequent PR.

parity/rpc_apis.rs
vorot93: It's been in this state for years already - so making an even bigger warning is just kicking this can down the road.

PR #11, opened 2020-08-05
Bugfixes (#11400) (#11444)

rakita:
+ Avoid long state queries when serving GetNodeData requests (#11444)
+ verification: fix race same block (#11400)

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1536,8 +1536,12 @@ impl ImportBlock for Client {
 				Ok(hash)
 			},
 			// we only care about block errors (not import errors)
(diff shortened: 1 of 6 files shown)


adria0:
The #11400 is about the GetNodeData call, but AFAIK are we going to remove this call eventually, isn't it? @vorot93 @rakita

vorot93:
Yes, our peers do not benefit from and should not participate in fast sync. I suggest cutting #11444 from this PR, and making a clean GetNodeData removal PR instead.

rakita:
I created two separate branches. One that removed GeNodeData and one that fixes 11444. I still need to test them and open new pull request. I am closing this.

PR #20, opened 2020-08-08
Remove private transactions, consensus messages, AuRa empty steps

vorot93:
Private transactions were deprecated in 3.0.

This also removes consensus messages and empty steps that depend on them. Therefore it's also a revert of #7605.


vorot93:
Closing as too ambitious for a single PR.

PR #23, opened 2020-08-10
Remove consensus messages, AuRa empty steps support

vorot93:
This removes consensus messages and AuRa empty steps that depend on them. Therefore it's also a revert of #7605.


rakita:
Decided not to remove currently

PR #27, opened 2020-08-17
Prometheus

rakita:


parity/export_hardcoded_sync.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
```
adria0, line 1: Is this related to Prometheus?

parity/configuration.rs
```diff
@@ -159,13 +166,14 @@ impl Configuration {
...
         let format = self.format()?;
+        let metrics_conf = self.metrics_config()?;
         let keys_iterations = NonZeroU32::new(self.args.arg_keys_iterations)
             .ok_or_else(|| "--keys-iterations must be non-zero")?;
 
         let cmd = if self.args.flag_version {
             Cmd::Version
         } else if self.args.cmd_signer {
-            let authfile = ::signer::codes_path(&ws_conf.signer_path);
+            let authfile = crate::signer::codes_path(&ws_conf.signer_path);
```
adria0, line 176: Is this related to Prometheus?

parity/rpc.rs
```diff
@@ -180,7 +182,7 @@ pub fn new_ws<D: rpc_apis::Dependencies>(
     let signer_path;
     let path = match conf.support_token_api {
         true => {
-            signer_path = ::signer::codes_path(&conf.signer_path);
+            signer_path = crate::signer::codes_path(&conf.signer_path);
```
adria0, line 185: Is this related to Prometheus?

parity/export_hardcoded_sync.rs
rakita: Dead code left after merge. Will be removed.

parity/configuration.rs
rakita: The answer is no, but it is just syntax change needed fo the rust 2018 edition.

parity/rpc.rs
rakita: The answer is no, but it is just syntax change needed fo the rust 2018 edition.

parity/configuration.rs
adria0: Oh, yes, I do not know that I was thinking... Lol

scripts/prometheus/docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+      dockerfile: scripts/docker/alpine/Dockerfile
+      context: ../..
+    ports:
+      - '30303:30303'
+      - '30303:30303/udp'
+      - '8545:8545'
+    links:
+      - prometheus
+
+    entrypoint: ["/home/openethereum/openethereum","--metrics","--metrics-interface=all"]
```
adria0, line 15:
```suggestion
    entrypoint: ["/home/parity/parity","--metrics","--metrics-interface=all"]
```

parity/rpc.rs
rakita: This is still valid with 2015

Cargo.toml
```diff
@@ -4,6 +4,7 @@ name = "parity-ethereum"
 # NOTE Make sure to update util/version/Cargo.toml as well
 version = "2.5.13"
 license = "GPL-3.0"
+edition = "2018"
```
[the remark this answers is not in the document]
rakita: Using edition 2018 is step forward and this does not change the version of the compiler we are using. I would like this to stay, but for the current state, I don't have a strong opinion about it.

util/runtime/Cargo.toml
```diff
@@ -7,5 +7,7 @@ version = "0.1.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 
 [dependencies]
-futures = "0.1"
-tokio = "0.1.22"
+futures03 = {package = "futures",  version = "0.3.5", default-features = false, features = ["compat"] }
+futures = { package = "futures", version = "0.1" }
+tokio-compat = "0.1.6"
+tokio = { version = "0.1.22"}
```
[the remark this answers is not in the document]
rakita: I already tested this and I feel like it is a waste to just remove it, cause we could use it in the future.

util/journaldb/src/overlayrecentdb.rs
```diff
@@ -288,15 +288,19 @@ impl JournalDB for OverlayRecentDB {
...
 
-    fn mem_used(&self) -> usize {
-        let mut mem = self.transaction_overlay.mem_used();
-        let overlay = self.journal_overlay.read();
-
-        mem += overlay.backing_overlay.mem_used();
-        mem += overlay.pending_overlay.heap_size_of_children();
-        mem += overlay.journal.heap_size_of_children();
+    fn get_sizes(&self, sizes: &mut BTreeMap<String, usize>) {
+        //sizes.insert(String::from("db_overlay_recent_transactions_size"), refs_size); //TODO MemoryDB
```
adria0, line 292: Could you descibe with more detail what is this TODO, please?

rakita: The current version of MemoryDB is just returning 0 on `mem_used()` and we don't have access to the underlying HashMap that is used in MemoryDB to at least get `len()`. The first intention for TODO is to change this when we update MemoryDB to new version, but now when I am thinking about it, we could just copy memory-db into project and add `lenI()` function.

adria0 (approved):

PR #28, opened 2020-08-20
Implement eth/64 (EIP-2364) and drop support for eth/62

adria0:
Backport of https://github.com/openethereum/openethereum/pull/11472
To me merged after https://github.com/vorot93/oe_playground/pull/26


ethcore/sync/src/chain/handler.rs
```diff
@@ -659,9 +659,22 @@ impl SyncHandler {
...
         sync.handshaking_peers.remove(&peer_id);
         let protocol_version: u8 = r.val_at(0)?;
+        let eth_protocol_version = io.protocol_version(&ETH_PROTOCOL, peer_id);
         let warp_protocol_version = io.protocol_version(&PAR_PROTOCOL, peer_id);
         let warp_protocol = warp_protocol_version != 0;
         let private_tx_protocol = warp_protocol_version >= PAR_PROTOCOL_VERSION_3.0;
+
+        let forkid_validation_error = {
+            if eth_protocol_version >= ETH_PROTOCOL_VERSION_64.0 {
+                let fork_id = rlp04::Rlp::new(r.as_raw()).val_at(5)?;
```
rakita, line 669: If you take the value at index 5 you need to move indexes bellow by one (line 698 and others will not be valid). If it is not protocol 64 you leave it as it is.
In 3.0.1 this is done by iterator but it can be done with index increment.

adria0: Oh, yes. Good catch.

adria0: Done, please check https://github.com/vorot93/oe_playground/pull/28/commits/cd554547b1c1823e432b417a6dc0217b949b74ee

rakita (approved):
lgtm

PR #31, opened 2020-08-21
Rebranding to OpenEthereum

rakita:
* Legal headers
* Network ID change
* binary file change
* Links to openethereum github


ethcore/src/lib.rs
```diff
@@ -1,62 +1,22 @@
...
-//!
-//!   # download and build parity
-//!   git clone https://github.com/paritytech/parity-ethereum
-//!   cd parity
-//!   cargo build --release
-//!   ```
-
-// Recursion limit required because of
-// error_chain foreign_links.
-#![recursion_limit = "128"]
```
adria0, line 59: Is this not still used?

Cargo.toml
```diff
@@ -1,10 +1,13 @@
 [package]
-description = "Parity Ethereum client"
-name = "parity-ethereum"
+description = "OpenEthereum"
+name = "openethereum"
 # NOTE Make sure to update util/version/Cargo.toml as well
-version = "2.5.13"
+version = "3.0.1"
```
adria0, line 5: 3.1 ?

scripts/docker/alpine/Dockerfile
```diff
@@ -10,34 +10,37 @@ RUN apk add --no-cache \
...
 # show backtraces
 ENV RUST_BACKTRACE 1
 
+# curl and jq are installed to help create health and readiness checks on Kubernetes
 RUN apk add --no-cache \
   libstdc++ \
   eudev-libs \
-  libgcc
+  libgcc \
+  curl \
```
adria0, line 31: Is this a backport of other issue?

ethcore/src/lib.rs
rakita: The default value is 128, this is why i omitted it. https://doc.rust-lang.org/reference/attributes/limits.html

adria0 (approved):
LGTM

Cargo.toml
rakita: Not sure what is practice, do we update to 3.1.0 just before release or we are putting the next version after release. I would put 3.1 as last commit before our release, but for version, I don't have a strong opinion.

scripts/docker/alpine/Dockerfile
rakita: I tried clang and it fails to compile kvdb this is why I left gcc. 

For curl it is used by Kubernetes or at least it says that in newest docker file: `# curl and jq are installed to help create health and readiness checks on Kubernetes`

ethcore/src/lib.rs
adria0: > The default value is 128, this is why i omitted it. https://doc.rust-lang.org/reference/attributes/limits.html

Good catch!

PR #33, opened 2020-09-01
Remove accounts bloom

vorot93:
Backport of #11589.


adria0 (commented):
I've been doing this change in a local branch with the database upgrade, and I found that the `ethcore_bloom_journal` crate is not used anywhere, so it should be removed, also.

ethcore/db/src/db.rs
```diff
@@ -33,8 +33,6 @@ pub const COL_BODIES: Option<u32> = Some(2);
 pub const COL_EXTRA: Option<u32> = Some(3);
 /// Column for Traces
 pub const COL_TRACE: Option<u32> = Some(4);
-/// Column for the empty accounts bloom filter.
-pub const COL_ACCOUNT_BLOOM: Option<u32> = Some(5);
```
adria0, line 37: When doing database migration, the data of this column should be deleted (not the column itself), so is good to keep the column, only for this purpose, I put the name `COL_ACOUNT_BLOOL_UNUSED`

vorot93: I was hesitant about backporting the migration as it could break 2.5 backwards compatibility, and if the accounts bloom savings are worth it (it's really tiny AFAIK).

ethcore/db/src/db.rs
```diff
@@ -33,8 +33,6 @@ pub const COL_BODIES: Option<u32> = Some(2);
...
 /// Column for general information from the local node which can persist.
 pub const COL_NODE_INFO: Option<u32> = Some(6);
 /// Number of columns in DB
```
adria0, line 38: `pub const NUM_COLUMNS: Option<u32> = Some(7);` ?

adria0:
In `parity/db/rocksdb/migration.rs`, update database model to 16

```
/// Current version of database models.
const CURRENT_VERSION: u32 = 16;
```

rakita (approved):
LGTM, just address Adria comment before merge.

adria0 (approved):
LGTM

PR #34, opened 2020-09-01
Remove account management

vorot93:
This PR removes account management facilities from OpenEthereum. They were deprecated around two years ago - the rationale and discussion can be found in issue 9997.

List of changes:
- CLIs for managing accounts and WebSocket-based signing API have been removed.
- `parity_accounts` and `personal` RPC APIs have been removed.
- Secret Store now accepts raw private key instead of account address and password.
- Private key for PoA signers can be set via `ENGINE_SIGNER` environment variable on startup, as well as through `parity_setEngineSignerSecret` RPC call dynamically.

To be merged after #22.


gituser:
Don't remove this please.
A lot of people still depend on account management.

Even geth still have it and it's very useful!

Proof:

```
$ ./geth --version
Incorrect Usage. flag provided but not defined: -version

NAME:
   geth - the go-ethereum command line interface

   Copyright 2013-2019 The go-ethereum Authors

USAGE:
   geth [options] command [command options] [arguments...]
   
VERSION:
   1.9.20-stable-979fc968
   
COMMANDS:
   account                            Manage accounts
```

```
$ ./geth account
NAME:
   geth account - 

Manage accounts, list all existing accounts, import a private key into a new
account, create a new account or update an existing account.

It supports interactive mode, when you are prompted for password as well as
non-interactive mode where passwords are supplied via a given password file.
Non-interactive mode is only meant for scripted use on test networks or known
safe environments.

Make sure you remember the password you gave when creating a new account (with
either new or import). Without it you are not able to unlock your account.

Note that exporting your key in unencrypted format is NOT supported.

Keys are stored under <DATADIR>/keystore.
It is safe to transfer the entire directory or the individual keys therein
between ethereum nodes by simply copying.

Make sure you backup your keys regularly.

USAGE:
   geth account command [command options] [arguments...]

COMMANDS:
     list                             Print summary of existing accounts
     new                              Create a new account
     update                           Update an existing account
     import                           Import a private key into a new account

OPTIONS:
   --help, -h                         show help
```

vorot93:
@gituser The *only* case where there is necessary is a PoA node signing operator - for this there will be an environment variable where the private key can be loaded.

Other than that there is *no* reason why we should deal with private keys.

gituser:
If you check the mentioned issue https://github.com/openethereum/openethereum/issues/9997 you'll see the discussion as there is no real alternative out there to replace this mechanism.

* clef seems to be needing some sort of interaction and hasn't been really tested with parity
* no real alternatives at the moment for account management except building your own and it might be tricky for many users as it's basic functionality to send a transaction within a client
* geth didn't remove account management either
* removing something basic from the client will kill the client - more complexity: to send transaction you need to install third-party tools and/or code your own. 
* there are many cases where it's more useful to use built-in account management alongside with parity_* RPCs to get mempool pending transaction via parity_localTransactions, to track what is pending instead of scanning whole blockchain
* why remove/break what is working for many people?

vorot93:
@gituser

> clef seems to be needing some sort of interaction and hasn't been really tested with parity

Then feel free to test and report bugs. Quick Google tells me there are also other offline wallets available.
> no real alternatives at the moment for account management except building your own and it might be tricky for many users as it's basic functionality to send a transaction within a client

We do allow sending *signed* transactions from the client.
> geth didn't remove account management either

It's their decision, we don't discuss what other clients do or don't do.
> removing something basic from the client will kill the client - more complexity: to send transaction you need to install third-party tools and/or code your own.

And it *is* the right way to manage private keys. Separate from devp2p network, in an isolated node, or even secure enclave. We don't want to bear this responsibility
> there are many cases where it's more useful to use built-in account management alongside with parity_* RPCs to get mempool pending transaction via parity_localTransactions, to track what is pending instead of scanning whole blockchain

`--tx-queue-locals`
> why remove/break what is working for many people?

It's only working until something wrong happens with the client or the environment where the client is run. This is a huge codebase as it is, and we don't have the resources to continuously audit account management. Add to that the fact that it's just bad security practice - and here you get the grounds for removal.

gituser:
@vorot93 

>Then feel free to test and report bugs. Quick Google tells me there are also other offline wallets available.

The problem it's a huge effort for operators / services to "google and find solutions" when everything works at the moment or implement inside their own stack. 

Would be better if you listed actually here available solutions, "drop-in" replacements for RPCs that you want to entirely remove before doing so.

>And it is the right way to manage private keys. Separate from devp2p network, in an isolated node, or even secure enclave. We don't want to bear this responsibility

It depends. Bitcoin didn't remove account management and private keys, but you can disable it if you want to, just start bitcoin with `disablewallet=1` . I'd suggest doing the same for parity. Make an option to disable it if need to.

>It's only working until something wrong happens with the client or the environment where the client is run.
>This is a huge codebase as it is, and we don't have the resources to continuously audit account management. 

Crippling client down won't make it better instead of fixing actual issues.
Why do you need to continuously audit account management?

>Add to that the fact that it's just bad security practice - and here you get the grounds for removal.

I don't agree with this.

rakita:
Decided not to remove

gituser:
Great news, thank you very much!

PR #38, opened 2020-09-11
EIP-2464: eth/65: transaction announcements and retrievals

vorot93:


ethcore/sync/src/chain/mod.rs
```diff
@@ -769,10 +827,37 @@ impl ChainSync {
...
+        let imported = txs.iter().map(|tx| tx.hash()).collect::<H256FastSet>();
+        for (pid, peer_info) in &mut self.peers {
+            peer_info.unfetched_pooled_transactions = peer_info
+                .unfetched_pooled_transactions
+                .difference(&imported)
+                .copied()
+                .collect();
+            if *pid == peer_id {
+                match GetPooledTransactionsReport::generate(
+                    std::mem::replace(&mut peer_info.asking_pooled_transactions, Vec::new()),
```
tomusdrw, line 840: You might rather `peer_info.asking_pooled_transactions.drain()` here and change the signature of `generate` to take an `Iterator` as well to avoid re-allocations.

ethcore/sync/src/chain/mod.rs
```diff
@@ -769,10 +827,37 @@ impl ChainSync {
...
+                    txs.iter().map(UnverifiedTransaction::hash),
```
tomusdrw, line 841: We already do `hash`ing when we construct `imported`, couldn't that be re-used?

ethcore/sync/src/chain/mod.rs
```diff
@@ -675,6 +684,55 @@ pub struct ChainSync {
...
+    ) -> Result<Self, H256> {
+        let mut out = GetPooledTransactionsReport::default();
+
+        let asked_set = asked.iter().copied().collect::<H256FastSet>();
+        let mut asked_iter = asked.drain(std::ops::RangeFull);
+        let mut txs = received.into_iter();
+        let mut next_received: Option<H256> = None;
+        loop {
+            if let Some(received) = next_received {
+                if !asked_set.contains(&received) {
```
tomusdrw, line 707: Seems like this could be done with a set difference (`received - asked != 0`) more efficiently, afaict from the context where `generate` is being used, we do have a set of `imported` transactions already.

ethcore/sync/src/chain/mod.rs
```diff
@@ -675,6 +684,55 @@ pub struct ChainSync {
...
+                    return Err(received);
+                }
+
+                if let Some(requested) = asked_iter.next() {
```
tomusdrw, line 711: I guess we assume that both iterators have to be sorted, right? I'm not sure I understand the point of the entire loop then, if this again could be done with simple (and efficient) set operations.
We basically have two sets: `received` and `asked` and we need:
```
superfluous = received - asked;
found = received * asked;
missing = asked - found;
```

tomusdrw: Okay, seems we also want to split `missing` into `requested-but-missing` and `not_sent`. But that should be doable as well (given all the resulting sets are ordered: `not_sent = asked[last(found)...], missing = asked[...last(found)] - found`).

If you prefer a loop I also feel this could be simplified if we keep the collections ordered (unless it's not possible cause `H256` doesn't do `PartialOrd` - don't remember exactly).
```
let received : VecDeque<>;
let asked: Iterator<H256>; // ordered by `H256`
loop {
   match (asked.next(), received.pop_front()) {
     (Some(expected), Some(got)) if expected == got => {
          found.insert(expected);
     },
     (Some(expected), Some(got)) if expected < got => {
          missing.insert(expected);
          received.push_front(got);
     },
     (Some(expected), Some(got)) if expected > got => {
          return Err(got);
     },
     (None, Some(got)) => {
          return Err(got);
     },
     (Some(expected), None) => {
         not_sent.insert(expected); // we can early-exit here and just append the rest of the iterator.
     },
     (None, None) => break;
   }
}
return Ok(out);
```

ethcore/sync/src/chain/mod.rs
```diff
@@ -769,10 +827,37 @@ impl ChainSync {
...
-        if let Some(peer_info) = self.peers.get_mut(&peer_id) {
-            peer_info
-                .last_sent_transactions
-                .extend(txs.iter().map(|tx| tx.hash()));
+        // Remove imported txs from all request queues
+        let imported = txs.iter().map(|tx| tx.hash()).collect::<H256FastSet>();
+        for (pid, peer_info) in &mut self.peers {
+            peer_info.unfetched_pooled_transactions = peer_info
+                .unfetched_pooled_transactions
+                .difference(&imported)
```
tomusdrw, line 835: I think you could just use `remove` here to avoid copying and re-alloc?

ethcore/sync/src/chain/mod.rs
```diff
@@ -769,10 +827,37 @@ impl ChainSync {
...
+                match GetPooledTransactionsReport::generate(
+                    std::mem::replace(&mut peer_info.asking_pooled_transactions, Vec::new()),
+                    txs.iter().map(UnverifiedTransaction::hash),
+                ) {
+                    Ok(report) => {
+                        // Some transactions were not received in this batch because of size.
+                        // Add them back to request feed.
+                        peer_info.unfetched_pooled_transactions = peer_info
+                            .unfetched_pooled_transactions
+                            .union(&report.not_sent)
```
tomusdrw, line 848: `.extend` instead of `union + copy`.

ethcore/sync/src/chain/mod.rs
```diff
@@ -769,10 +827,37 @@ impl ChainSync {
...
+                            .collect();
+                    }
+                    Err(unknown_tx) => {
+                        // punish peer?
+                    }
+                }
+
+                peer_info
+                    .last_sent_transactions
+                    .extend(txs.iter().map(|tx| tx.hash()));
```
tomusdrw, line 859: Hashing once again? I really hope it's cached :)

ethcore/sync/src/chain/propagator.rs
```diff
@@ -270,7 +297,7 @@ impl SyncPropagator {
                 .chain(&to_send)
                 .cloned()
                 .collect();
-            send_packet(io, peer_id, to_send.len(), packet.out());
+            send_packet(io, peer_id, is_hashes, to_send.len(), packet.out());
```
tomusdrw, line 300: Hmm, `last_sent_transactions` and `stats.propagated` will now think that the peers knows about these transactions already, but in fact it only knows about the hashes. I supposed it's good enough though, but might mess up some statistics.

ethcore/sync/src/chain/supplier.rs
```diff
@@ -268,6 +268,30 @@ impl SyncSupplier {
...
+    /// Respond to GetPooledTransactions request
+    fn return_pooled_transactions(io: &dyn SyncIo, r: &Rlp, peer_id: PeerId) -> RlpResponseResult {
+        let payload_soft_limit = io.payload_soft_limit();
+
+        let mut added = 0;
+        let mut rlp = RlpStream::new();
+        rlp.begin_unbounded_list();
+        for v in r {
+            if let Ok(hash) = v.as_val::<H256>() {
+                if let Some(tx) = io.chain().queued_transaction(hash) {
```
tomusdrw, line 280: The transaction might be in a block already, but I suppose it's fine, since the peer should get the same block soon as well and they don't really need the transaction in the pool any more.

ethcore/sync/src/chain/handler.rs
```diff
@@ -811,6 +816,59 @@ impl SyncHandler {
...
+    ) -> Result<(), DownloaderImportError> {
+        for item in tx_rlp {
+            let hash = item
+                .as_val::<H256>()
+                .map_err(|_| DownloaderImportError::Invalid)?;
+
+            if io.chain().queued_transaction(hash).is_none() {
+                sync.peers
+                    .get_mut(&peer_id)
+                    .map(|peer| peer.unfetched_pooled_transactions.insert(hash));
```
tomusdrw, line 834: There should be some reasonable upper bound here.
1. We don't verify the peer speaks 65 at all, when we receive a package. So it may inform us about new hashes, but we never really check them.
2. They can keep spamming random hashes to us, and we will figure that we don't have them in the pool and keep extending this set.

rakita:
We decided to not include this change into 3.1 cause it is not a requirement for berlin fork

PR #42, opened 2020-09-16
Remove pwasm and unused EIPs

vorot93:


rakita:
Decided not to remove

PR #51, opened 2020-09-21
Bump jsonrpc-* to v15 into main

adria0:


rakita (approved):
LGTM. But when I build something I know there is Cargo.lock conflict I want to minimize its change. I do merge, checkout Cargo.lock from main, do build with this Cargo.lock and only minimal changes are made. And from time to time, we can make a clean Build without lock just to update everything.

PR #60, opened 2020-09-25
Add `wasmDisableTransition` spec option

varasev:
Adds `wasmDisableTransition` spec option to disable WASM on Kovan on a certain block.

POA Team will need this to avoid splitting the Kovan chain after some Kovan validators replace their node with Nethermind client (which doesn't support WASM).


tkstanczak:
looks ok

rakita (approved):
lgtm

adria0 (approved):
LGTM

PR #74, opened 2020-10-16
Sync block verification

rakita:
- Parallelization of verification that depends on a parent-child relationship can be troublesome this commit removes that and makes that part of the verification sequential.
- Batch executing one by one block was faster on my machine and it made the client more stable.


adria0 (approved):
So, mainly this is, in `import_verified_blocks`

- Do not resignal block verification *before* importing blocks, do it *after* importing blocks. This is done by removing the resignaling from `drain`.
-  Change `max_round_blocks_to_import` from `12` to `1`

LGTM

PR #76, opened 2020-10-21
Transaction Type and EIP-2930

rakita:


rakita:
It seems that build fails with msg ```can't find crate for `serde_derive` which `rustc_version` depends on```
Some other users can this problem with actions/cache@v2 reported at https://github.com/actions-rs/cargo/issues/111 because of actiong/cache@2 . Possible solution is to do something like `cargo clean --locked --release -p serde_derive` found here: https://github.com/denoland/deno/commit/7b3786983598c30a08b591c36f34b43c7aeec379

ethcore/src/executive.rs
```diff
@@ -45,6 +44,11 @@ const UNPRUNABLE_PRECOMPILE_ADDRESS: Option<Address> = Some(ethereum_types::H160
 /// Precompile that can never be prunned from state trie (none)
 const UNPRUNABLE_PRECOMPILE_ADDRESS: Option<Address> = None;
 
+/// Gas per received storage key
```
adria0, line 47: What about defining the value in `Schedule` https://github.com/openethereum/backport-3.1/blob/rakita/transaction_type/ethcore/vm/src/schedule.rs ?

ethcore/src/spec/spec.rs
```diff
@@ -137,6 +137,8 @@ pub struct CommonParams {
     pub eip2315_transition: BlockNumber,
     /// Number of first block where EIP-2929 rules begin.
     pub eip2929_transition: BlockNumber,
+    /// Number of first block where EIP-2929 rules begin.
```
adria0, line 140:
```suggestion
    /// Number of first block where EIP-2930 rules begin.
```

ethcore/src/executive.rs
```diff
@@ -1136,16 +1144,50 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
         T: Tracer,
         V: VMTracer,
     {
+        let schedule = self.schedule;
+
+        //check if particualar transaction type is enabled at this block number in schedule
+        match t.as_unsigned() {
+            TypedTransaction::AccessList(_) => {
+                if !(schedule.eip2929 && schedule.eip2930) {
```
adria0, line 1152:
```suggestion
                if !schedule.eip2930 {
```

In general, I do not see in other parts of the code checking EIP dependencies, so proballly it will consistent only checking for `eip2930`.

rpc/src/v1/types/transaction.rs
```diff
@@ -67,6 +69,10 @@ pub struct Transaction {
     pub s: U256,
     /// Transaction activates at specified block.
     pub condition: Option<TransactionCondition>,
+    /// transaction type
+    pub tx_type: u8,
+    /// optional access list
+    pub optional_access_list: Option<Vec<(H160, Vec<H256>)>>,
```
adria0, line 75:
```suggestion
    pub access_list: Vec<(H160, Vec<H256>)>,
```

I do not see using `optional_` prefix in other parts of the code for vec, and in EIP2929 the list is named `access_list`. So I propose to name it `access_list`.

Since `Vec::new()` does not allocate until any element is inserted  (https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.new), makes sense to be an `Option`?

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +212,134 @@ impl HeapSizeOf for Transaction {
...
+    }
+
+    // encode by this payload spec: 0x03 | rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])
+    pub fn encode(&self, signature: Option<&SignatureComponents>) -> Vec<u8> {
+        let mut stream = RlpStream::new();
+        //stream.begin_list(2);
+        //stream.append(&3u8);
+
+        let mut list_size = 7;
+        list_size += if signature.is_some() { 3 } else { 0 };
```
adria0, line 291:
```suggestion
        let list_size = if signature.is_some {
            10
        } else {
            7
        }
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +212,134 @@ impl HeapSizeOf for Transaction {
...
+        let accl_rlp = tx_rlp.at(6)?;
+
+        // access_list pattern: [[{20 bytes}, [{32 bytes}...]]...]
+        let mut accl: Vec<(H160, Vec<H256>)> = Vec::new();
+
+        for i in 0..accl_rlp.item_count()? {
+            let accounts = accl_rlp.at(i)?;
+
+            if accounts.item_count()? != 2 {
+                //TODO check what to do if we have only one item, should we be strict or not
```
adria0, line 259: I think that this is ok, in EIP2929 `process_access_list` pseudocode says

`assert isinstance(item, list) and len(item) == 2`

ethcore/types/src/transaction/transaction.rs
```diff
@@ -219,22 +354,11 @@ impl Transaction {
     pub fn with_signature(self, sig: Signature, chain_id: Option<u64>) -> UnverifiedTransaction {
```
adria0, line 354: Reading EIP2718, it seems that typed transactions are designed to potentially be signatureless (_Another potential situation is one where transactions don’t have a signature in the traditional sense and instead have some other mechanism for determining validity._), so maybe signature methods `with_signature` or `null_sign`  belongs to `Transaction`, not to `TypedTransaction`

adria0: Nonetheless, properly implementing this abstraction also breaks the current `SignedTransaction`/`UnverifiedTransaction`, just for something in the future that we do not know. So for me it's ok to leave as it is now.

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +212,134 @@ impl HeapSizeOf for Transaction {
...
 
-impl Transaction {
     /// The message hash of the transaction.
     pub fn hash(&self, chain_id: Option<u64>) -> H256 {
-        let mut stream = RlpStream::new();
-        self.rlp_append_unsigned_transaction(&mut stream, chain_id);
-        keccak(stream.as_raw())
+        match self {
+            Self::Legacy(tx) => tx.hash(chain_id),
+            Self::AccessList(ocl) => ocl.hash(),
```
adria0, line 341:
```suggestion
            Self::AccessList(ocl) => ocl.hash(chain_id),
```

I think that is not explicity specified in 2930, but https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md also applies, that is adding the `chain_id` in the RLP only when veryfiying the signature.

ethcore/src/executive.rs
rakita: Moved to Schedule

ethcore/src/executive.rs
rakita: changed to eip2930 only

rpc/src/v1/types/transaction.rs
rakita: Agreed

ethcore/types/src/transaction/transaction.rs
rakita: This makes more sense now that we include chain_id.

ethcore/types/src/transaction/transaction.rs
rakita: Agreed, removing the comment

ethcore/src/machine/impls.rs
```diff
@@ -456,16 +455,15 @@ impl EthereumMachine {
         &self,
         transaction: &[u8],
     ) -> Result<UnverifiedTransaction, transaction::Error> {
-        let rlp = Rlp::new(&transaction);
-        if rlp.as_raw().len() > self.params().max_transaction_size {
+        if transaction.len() > self.params().max_transaction_size {
```
adria0, line 458: nice change

ethcore/src/state/mod.rs
```diff
@@ -954,8 +955,11 @@ impl<B: Backend> State<B> {
             TransactionOutcome::StateRoot(self.root().clone())
         };
 
-        let output = e.output;
-        let receipt = Receipt::new(outcome, e.cumulative_gas_used, e.logs);
+        let output = e.output; //TODO dr create new types of receipts
```
adria0, line 958: This TODO must be removed or kept?

ethcore/types/src/receipt.rs
```diff
@@ -60,10 +64,32 @@ impl Receipt {
             outcome,
         }
     }
-}
+    pub fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
+        if rlp.item_count()? == 3 {
```
adria0, line 68: `if rlp.item_count()? == 0` could crash?

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+
+    pub fn tx(&self) -> &Transaction {
+        &self.transaction
+    }
+
+    pub fn tx_mut(&mut self) -> &mut Transaction {
+        &mut self.transaction
+    }
+
+    // decode bytes by this payload spec: rlp([3, [nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])
```
adria0, line 265:
```suggestion
    // decode bytes by this payload spec: rlp([3, [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]])
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+    pub fn decode(tx: &[u8]) -> Result<UnverifiedTransaction, DecoderError> {
+        let tx_rlp = &Rlp::new(tx);
+
+        // we need to have 10 items in this list
```
adria0, line 269:
```suggestion
        // we need to have 11 items in this list
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+
+        // access_list pattern: [[{20 bytes}, [{32 bytes}...]]...]
+        let mut accl: AccessList = Vec::new();
+
+        for i in 0..accl_rlp.item_count()? {
+            let accounts = accl_rlp.at(i)?;
+
+            // check if there is list of 2 items
+            if accounts.item_count()? != 2 {
+                return Err(DecoderError::Custom("Unknown access list lenght"));
```
adria0, line 291:
```suggestion
                return Err(DecoderError::Custom("Unknown access list length"));
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+        let signature = SignatureComponents {
+            v: tx_rlp.val_at(8)?,
+            r: tx_rlp.val_at(9)?,
+            s: tx_rlp.val_at(10)?,
         };
-        match secret {
-            Some(s) => tx.sign(&s, None),
-            None => tx.null_sign(1),
+
+        //and here we create UnverifiedTransaction and calculate its hash
```
adria0, line 303:
```suggestion
        // and here we create UnverifiedTransaction and calculate its hash
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+    ) -> Vec<u8> {
+        let mut stream = RlpStream::new();
+
+        let list_size = if signature.is_some() { 11 } else { 8 };
+        stream.begin_list(list_size);
+
+        // append chain_id. from EIP-2930: chainId is defined to be an integer of arbitrary size.
+        stream.append(&(if let Some(n) = chain_id { n } else { 0 }));
+
+        //append legacy transaction
```
adria0, line 330:
```suggestion
        // append legacy transaction
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+        self.transaction.rlp_append_open(&mut stream);
+
+        //access list
```
adria0, line 333:
```suggestion
        // access list
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
+                stream.append(storage_key);
+            }
+        }
+
+        // append signature
+        if let Some(signature) = signature {
+            signature.rlp_append(&mut stream);
         }
+
+        //make as vector of bytes
```
adria0, line 349:
```suggestion
        // make as vector of bytes
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -152,59 +235,157 @@ impl HeapSizeOf for Transaction {
...
-            v: t.v.into(),
-            hash: 0.into(),
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum TypedTransaction {
+    Legacy(Transaction), // old legacy RLP encoded transaction
+    AccessList(AccessListTx), // EIP-2930 Transaction with a list of addresses and storage keys that the transaction plans to access.
+                              // Accesses outside the list are possible, but become more expensive.
+}
+
+//Function that are batched from Transaction struct and needs to be reimplemented
```
adria0, line 374: Should be this comment removed?

ethcore/types/src/transaction/transaction.rs
```diff
@@ -245,9 +416,12 @@ impl Transaction {
...
             transaction: UnverifiedTransaction {
                 unsigned: self,
-                r: U256::one(),
-                s: U256::one(),
-                v: 0,
+                chain_id: None,
+                signature: SignatureComponents {
+                    r: U256::one(),
+                    s: U256::one(),
+                    v: 4,
```
adria0, line 423: Why this `4` here?

ethcore/types/src/transaction/transaction.rs
```diff
@@ -259,35 +433,204 @@ impl Transaction {
     /// Legacy EIP-86 compatible empty signature.
     /// This method is used in json tests as well as
     /// signature verification tests.
-    #[cfg(any(test, feature = "test-helpers"))]
+    //#[cfg(any(test, feature = "test-helpers"))]
```
adria0, line 436: Should be this uncommented or removed?

ethcore/types/src/transaction/transaction.rs
```diff
@@ -259,35 +433,204 @@ impl Transaction {
...
+    fn decode_new(tx: &[u8]) -> Result<UnverifiedTransaction, DecoderError> {
+        if tx.is_empty() {
+            // at least one byte needs to be present
+            return Err(DecoderError::RlpIncorrectListLen);
+        }
+        let id = tx[0].try_into();
+        if id.is_err() {
+            return Err(DecoderError::Custom("Unknown transaction"));
+        }
+        //other transaction types
```
adria0, line 496:
```suggestion
        // other transaction types
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -259,35 +433,204 @@ impl Transaction {
...
+        }
+    }
+
+    pub fn decode(tx: &[u8]) -> Result<UnverifiedTransaction, DecoderError> {
+        if tx.is_empty() {
+            // at least one byte needs to be present
+            return Err(DecoderError::RlpIncorrectListLen);
+        }
+        let header = tx[0];
+        //type of transaction can be obtained from first byte. If first bit is 1 it means we are dealing with RLP list.
```
adria0, line 509:
```suggestion
        // type of transaction can be obtained from first byte. If first bit is 1 it means we are dealing with RLP list.
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -259,35 +433,204 @@ impl Transaction {
...
+        //if it is 0 it means that we are dealing with custom transaction defined in EIP-2718.
```
adria0, line 510:
```suggestion
        // if it is 0 it means that we are dealing with custom transaction defined in EIP-2718.
```

ethcore/types/src/transaction/transaction.rs
```diff
@@ -308,87 +651,71 @@ impl Deref for UnverifiedTransaction {
...
-        s.append(&self.v);
-        s.append(&self.r);
-        s.append(&self.s);
+        self.signature.r.is_zero() && self.signature.s.is_zero()
     }
 
     ///	Reference to unsigned part of this transaction.
-    pub fn as_unsigned(&self) -> &Transaction {
+    pub fn as_unsigned(&self) -> &TypedTransaction {
+        //TODO check where this is used
```
adria0, line 693: Keep or remove?

ethcore/types/src/transaction/transaction.rs
```diff
@@ -259,35 +433,204 @@ impl Transaction {
...
+/// Components that constitute transaction signature
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub struct UnverifiedTransaction {
-    /// Plain Transaction.
-    unsigned: Transaction,
+pub struct SignatureComponents {
     /// The V field of the signature; the LS bit described which half of the curve our point falls
     /// in. The MS bits describe which chain this transaction is for. If 27/28, its for all chains.
-    v: u64,
+    v: u8,
```
adria0, line 554: to avoid futher confustions, I would prefer to rename `v` to `stadard_v` or alike

ethcore/types/src/transaction/transaction.rs
```diff
@@ -450,7 +777,7 @@ impl HeapSizeOf for SignedTransaction {
 
 impl rlp::Encodable for SignedTransaction {
     fn rlp_append(&self, s: &mut RlpStream) {
-        self.transaction.rlp_append_sealed_transaction(s)
+        self.transaction.rlp_append(s)
```
adria0, line 780: +1

ethcore/types/src/transaction/transaction.rs
```diff
@@ -705,26 +1033,78 @@ mod tests {
...
+    #[test]
+    fn should_agree_with_geth_test() {
+        use rustc_hex::FromHex;
+        let encoded_tx = "01f8630103018261a894b94f5374fce5edbc8e2a8697c15331677e6ebf0b0a825544c001a0cb51495c66325615bcd591505577c9dde87bd59b04be2e6ba82f6d7bdea576e3a049e4f02f37666bd91a052a56e91e71e438590df861031ee9a321ce058df3dc2b";
+        let _ = TypedTransaction::decode(&FromHex::from_hex(encoded_tx).unwrap())
+            .expect("decoding tx data failed");
+    }
+
     #[test]
     fn should_agree_with_vitalik() {
```
adria0, line 1102: cool name for a test 🤣

ethcore/types/src/transaction/transaction.rs
rakita: I will use `standard_v`

ethcore/types/src/transaction/transaction.rs
rakita: It should represent an invalid signature. when `chain_id` and `standard_v` were merged this value was `0`. Why I use exactly `4` is because of [this ](https://github.com/openethereum/openethereum/blob/bf5830f766bf89861b92824de08f92a00de79d44/ethcore/types/src/transaction/transaction.rs#L102) code that marks `standard_v` invalid as `4`.

rakita: Because of this `4` i needed to change transaction hashes of few tests.

ethcore/types/src/receipt.rs
rakita: for length `1` it will fail in `else` branch and that is okay. But for `rlp.item_count()? == 5` or if it greater than 5 it will pass but still be invalid. I think it is okay for us to be strict here and check the exact lenght.

adria0 (approved):
LGTM

PR #81, opened 2020-10-26
Send RLPx auth in EIP-8 format

vorot93:
[EIP-8 specified a new handshake format for RLPx back in 2015.](https://eips.ethereum.org/EIPS/eip-8) It is far more extendable and computationally lighter on the nodes and used throughout all major clients today.

The new handshake format is also supported by OpenEthereum when another node initiates the connection. Unfortunately, however, OpenEthereum itself connects using the old handshake, most likely to support `parity` versions released prior to 2017.

I don't think this is relevant anymore. This PR upgrades outbound handshake to comply with EIP-8 as well. This is important because as I am building [rust-devp2p](https://github.com/rust-ethereum/devp2p) I have to either support two handshake formats (which requires extra maintenance and is more computationally intensive), or drop OpenEthereum to rust-devp2p connectivity. More than that, OE is the last blocker to *all* clients dropping pre-EIP-8 handshake support.

### DIFF
--- a/util/network-devp2p/src/handshake.rs
+++ b/util/network-devp2p/src/handshake.rs
@@ -20,16 +20,15 @@ use ethkey::{
     crypto::{ecdh, ecies},
     recover, sign, Generator, KeyPair, Public, Random, Secret,
 };
-use hash::write_keccak;
 use host::HostInfo;
 use io::{IoContext, StreamToken};
 use mio::tcp::*;
 use network::{Error, ErrorKind};
 use node_table::NodeId;
 use parity_bytes::Bytes;
-use rand::random;
+use rand::{random, Rng};
 use rlp::{Rlp, RlpStream};
-use std::time::Duration;
+use std::{iter::repeat_with, time::Duration};
 
 #[derive(PartialEq, Eq, Debug)]
 enum HandshakeState {
@@ -39,10 +38,10 @@ enum HandshakeState {
     ReadingAuth,
     /// Waiting for extended auth packet
     ReadingAuthEip8,
-    /// Waiting for ack packet
-    ReadingAck,
-    /// Waiting for extended ack packet
-    ReadingAckEip8,
+    /// Waiting for ack packet len
+    ReadingAckLen,
+    /// Waiting for ack packet payload
+    ReadingAckPayload,
     /// Ready to start a session
     StartSession,
 }
@@ -74,7 +73,7 @@ pub struct Handshake {
 }
 
 const V4_AUTH_PACKET_SIZE: usize = 307;
-const V4_ACK_PACKET_SIZE: usize = 210;
+const ACK_PREFIX_LEN: usize = 2;
 const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(5);
 const PROTOCOL_VERSION: u64 = 4;
 // Amount of bytes added when encrypting with encryptECIES.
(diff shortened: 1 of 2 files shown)


ordian (commented):
It does make sense to me to drop the old protocol if it simplifies and optimizes thing.

But make sure it doesn't break anything by running the branch with `-l network` logs and maybe add more tests from the spec.

util/network-devp2p/src/handshake.rs
```diff
@@ -314,27 +298,26 @@ impl Handshake {
...
-        }
-        let message = ecies::encrypt(&self.id, &[], &data)?;
-        self.auth_cipher = message.clone();
-        self.connection.send(io, message);
-        self.connection.expect(V4_ACK_PACKET_SIZE);
-        self.state = HandshakeState::ReadingAck;
+        let encoded = rlp
+            .out()
+            .into_iter()
+            .chain(repeat(0).take(100 + random::<usize>() % 200))
```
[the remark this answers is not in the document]
ordian: > range [100, 300]

if this is math notation, then it should be `rand::thread_rng().gen_range::<usize>(100, 301)`

util/network-devp2p/src/handshake.rs
```diff
@@ -262,37 +260,23 @@ impl Handshake {
         Ok(())
     }
 
-    /// Parse and validate ack message
-    fn read_ack(&mut self, secret: &Secret, data: &[u8]) -> Result<(), Error> {
-        trace!(target: "network", "Received handshake ack from {:?}", self.connection.remote_addr_str());
-        if data.len() != V4_ACK_PACKET_SIZE {
-            debug!(target: "network", "Wrong ack packet size");
+    fn read_ack_len(&mut self, data: &[u8]) -> Result<(), Error> {
+        if data.len() != 2 {
```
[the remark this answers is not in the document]
vorot93: done

util/network-devp2p/src/handshake.rs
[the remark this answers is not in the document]
vorot93: done

vorot93:
@ordian Not sure about additional tests, but connecting to ethereum-sentry works now.

util/network-devp2p/src/handshake.rs
```diff
@@ -314,27 +311,26 @@ impl Handshake {
...
-        }
-        let message = ecies::encrypt(&self.id, &[], &data)?;
-        self.auth_cipher = message.clone();
-        self.connection.send(io, message);
-        self.connection.expect(V4_ACK_PACKET_SIZE);
-        self.state = HandshakeState::ReadingAck;
+        let encoded = rlp
+            .out()
+            .into_iter()
+            .chain(repeat_with(rand::random).take(rand::thread_rng().gen_range::<usize>(100, 301)))
```
ordian, line 324: actually looking at what geth does (although my `go` skills are weak), it seems to be zeroes in range `[100, 300)`
https://github.com/ethereum/go-ethereum/blob/97fc1c3b1d054a1345934e18fa6c3489e9119666/p2p/rlpx/rlpx.go#L592

vorot93: @ordian True, but implementations shouldn't read or touch this padding anyway.

fjl: It's OK to have random padding, but the padding will be encrypted, so it's also OK (and cheaper) to use zeros instead.

vorot93: Changed to zeroes

util/network-devp2p/src/handshake.rs
```diff
@@ -314,27 +311,26 @@ impl Handshake {
...
-        }
-        let message = ecies::encrypt(&self.id, &[], &data)?;
-        self.auth_cipher = message.clone();
-        self.connection.send(io, message);
-        self.connection.expect(V4_ACK_PACKET_SIZE);
-        self.state = HandshakeState::ReadingAck;
+        let encoded = rlp
+            .out()
+            .into_iter()
+            .chain(repeat(0).take(rand::thread_rng().gen_range::<usize>(100, 301)))
```
ordian, line 324: Let's avoid premature pessimization and use a static array of zeroes and `Vec::extend` instead.
LLVM is pretty decent, but it's not magic: https://rust.godbolt.org/z/vjobKj.
The latter generates much less code and uses one single memset.

vorot93: Replaced with `Vec::resize`.

rakita:
Opened new pull request with modifications here: https://github.com/openethereum/openethereum/pull/287

PR #98, opened 2020-11-13
Feature/improve dockerhub deployment

giacomolicari:
Add docker specs for the release of tagged and latest versions, also add specs for nightly builds.


denisgranha (approved):

giacomolicari:
I see there's an error from CI regarding Rust stuff.
I would merge the PR anyway since the it doesn't modify and script affected by that error.

PR #100, opened 2020-11-16
Trace comment on new block inclusion

rakita:
Comment only commit.


rakita:
it matches to this: https://openethereum.github.io/Trace-NewBlock

adria0 (changes requested):
Some comments are marked for rustdoc (`///`) and anothers no (`//`), maybe better if all are in the same class.

rakita:
changed to `/// t_nb`.

rakita:
changes all to `//` because `///` make rust doc break.

adria0 (approved):

PR #106, opened 2020-11-17
Fix deprecated set-env declaration

giacomolicari:
This PR replaces the old and now deprecated way to set environment variables on the CI.

For further information:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


rakita:
There is one more file `scripts/actions/install-sccache.ps1` that uses set-env

rakita (approved):

Issue #85, opened 2020-11-10
WS Error: Exceeded maximum buffer capacity


mdben1247:
- **OpenEthereum version**: 3.1.0
- **Operating system**: Linux
- **Installation**: built from source
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

An error occours when sending requests with websockets and the response is large in size (such as querying some historic logs):

`WS Error <Io(Custom { kind: InvalidInput, error: "Exceeded maximum buffer capacity" })>`

I guess this is due to upgrade in jsonrpc-core to v15.0.0, which now has a default maximum payload of 5MB. Attached is a small patch that adds a new setting: ws-max-payload. Ideally this would be split to in/out max size, but this version does not support it yet, see paritytech/jsonrpc#594

[max_payload.zip](https://github.com/openethereum/openethereum/files/5517160/max_payload.zip)

rakita:
done at https://github.com/openethereum/openethereum/pull/112

PR #112, opened 2020-11-19
Add ws-max-paxload

mdben1247:
I made an issue #85 with patch attached. I've since learned that the proper github way would be to create a pull request. 

This adds an option to set `max-payload` for WS server. By setting a large enough `max-payload,` we can avoid errors with requests larger than the default 5MB:

`WS Error <Io(Custom { kind: InvalidInput, error: "Exceeded maximum buffer capacity" })>`

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -33,35 +33,35 @@ jobs:
           profile:              minimal
           override:             true
       - name:                   Cache cargo registry
-        uses:                   actions/cache@v2
+        uses:                   actions/cache@v1.1.2
         with:
           path:                 ~/.cargo/registry
           key:                  ${{ runner.os }}-cargo-registry-build-tests-${{ hashFiles('**/Cargo.lock') }}
       - name:                   Cache cargo index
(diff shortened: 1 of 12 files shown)


adria0:
@mdben1247 could you set the merge target into `dev`, please?

mdben1247:
I changed target to dev, but now there are other file changes and commits listed, beside my fix. I am not very familiar with git workflow. Should I make another branch off of dev and cherry-pick my commit there?

adria0:
> I changed target to dev, but now there are other file changes and commits listed, beside my fix. I am not very familiar with git workflow. Should I make another branch off of dev and cherry-pick my commit there?

As you want, I usually merge & fix the conflicts.

mdben1247:
I leave it as is, since it seems that I'd have to open another pull request for that.

rakita:
I rebased it in here https://github.com/openethereum/openethereum/pull/155.
This pull request can be closed.

roninkaizen:
need a description usage example, please :)
going in advance with @itchykitteh, maybe it works even without knowing which settings are allowed

PR #125, opened 2020-11-24
fix: remove experimental from eth_getProof

gakonst:
Since Geth and Nethermind do not have the experimental flag for this, and given that the EIP has had no change for a while, should we remove this API call from being experimental?

Ref:
- https://github.com/ethereum/EIPs/issues/1186
- https://github.com/NethermindEth/nethermind/search?q=experimental
- https://github.com/ethereum/go-ethereum/search?q=experimental

### DIFF
--- a/rpc/src/v1/impls/eth.rs
+++ b/rpc/src/v1/impls/eth.rs
@@ -680,11 +680,6 @@ where
         values: Vec<H256>,
         num: Option<BlockNumber>,
     ) -> BoxFuture<EthAccount> {
-        try_bf!(errors::require_experimental(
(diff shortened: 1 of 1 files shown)


adria0:
@gakonst, parity developers tagged this feature as experimental, so I understand this `experimental` keyword as _not fully tested to be promoted to stable_. I think that if we remove the experimental tag, users will understand that this feature is fully tested and stable, and will start to use it in production. And I do not know if this is the case now.

rakita:
It seems geth is having this behind an `experimental` flag, and EIP is still not be finalized until then I think it is okay to stay as it is.

PR #138, opened 2020-12-01
Remove gitaction cache

rakita:
It happened again that the build fails because of gitaction cache, I think it is best to just disable it and run tests without it.


adria0 (approved):
LGTM

PR #149, opened 2020-12-04
Ancient target set. InvalidStateRoot bug (#69)

rakita:
#69

### DIFF
--- a/ethcore/sync/src/block_sync.rs
+++ b/ethcore/sync/src/block_sync.rs
@@ -516,8 +516,24 @@ impl BlockDownloader {
                         self.last_imported_hash
                     );
                 } else {
-                    let best = io.chain().chain_info().best_block_number;
-                    let best_hash = io.chain().chain_info().best_block_hash;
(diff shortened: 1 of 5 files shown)


ryanobjc:
This didn't seem to fix the bug I was experiencing.

rakita:
@ryanobjc Already corrupted DB is in most cases not possible to salvage. Removing DB and starting to sync again is probable what you need to do.

ethcore/sync/src/block_sync.rs
```diff
@@ -516,8 +516,24 @@ impl BlockDownloader {
...
+                                io.chain().chain_info().ancient_block_hash,
+                            ) {
+                                (best, best_hash) // best ancient block number and hash.
+                            } else {
+                                // None on ancient block/hash means that all ancient are already downloaded and stored in DB.
+                                self.state = State::Complete;
+                                return;
+                            }
+                        }
+                    };
```
rakita, line 536: This is fix

adria0 (approved):

PR #167, opened 2020-12-12
OpenEthereum rebranding

jmb-42:
Changed parity.service to openethereum.service to match what is listed in this documentation:

https://openethereum.github.io/FAQ#can-i-run-openethereum-as-a-daemon-or-system-service

I also updated a comment in the file about where to find the default config for users not running as root.


jmb-42:
I couldn't find a way to apply the `insubstantial` label, which I believe applies to this pr. I suspect that I don't have access to that functionality. I'm still familiarizing myself with the GitHub process, which is why I left this as a draft. Please let me know if I should have done anything different or what I can do next

rakita (approved):

PR #189, opened 2020-12-19
* Added checksum to the generated addresses

iosifpeterfi:
This commit adds checksum to the addresses generated by ethkey


rakita:
Not sure why this is needed.

Issue #169, opened 2020-12-14
Supplied instant is later than self


rakita:
- **OpenEthereum version**: v3.1.0 (probably this: 837e8b8725b87914ff43335d907411ad4c2a4669)
- **Operating system**: linux
- **Installation**: built from source
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

After client get synced to highest block and after ancient blocks start to download, client crashes with this msg:
```
eth1_1    | 2020-12-11 17:44:44 UTC Syncing #11432790 0x0589…bbd0     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+  252 Qed (Ancient:#5397346) LI:#11433033   48/50 peers      5 MiB chain   40 MiB queue  RPC:  0 conn,    2 req/s,  117 µs
eth1_1    | 2020-12-11 17:47:08 UTC Syncing #11432791 0x0c31…640b     0.01 blk/s    1.6 tx/s    0.1 Mgas/s      0+  260 Qed (Ancient:#5402621) LI:#11433041   51/50 peers     72 KiB chain   41 MiB queue  RPC:  0 conn,    0 req/s,   82 µs
eth1_1    | 2020-12-11 17:47:08 UTC Syncing #11432791 0x0c31…640b     0.01 blk/s    1.6 tx/s    0.1 Mgas/s      0+  260 Qed (Ancient:#5402621) LI:#11433041   51/50 peers    0 bytes chain   41 MiB queue  RPC:  0 conn,    0 req/s,   82 µs
eth1_1    | 2020-12-11 17:47:08 UTC Syncing #11432791 0x0c31…640b     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+  260 Qed (Ancient:#5402621) LI:#11433041   51/50 peers     72 KiB chain   41 MiB queue  RPC:  0 conn,    0 req/s,   82 µs
eth1_1    | 2020-12-11 17:47:08 UTC Syncing #11432791 0x0c31…640b     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+  260 Qed (Ancient:#5402621) LI:#11433041   51/50 peers     72 KiB chain   41 MiB queue  RPC:  0 conn,    0 req/s,   82 µs
[...]

PR #193, opened 2020-12-21
fix Supplied instant is later than self (#169)

rakita:
* Check in Informer periodic tick to skip all timer's calls that are called in succession. This is where the problem was.
* Added symbolic name in IoService thread pool. When crashes we could tell what pull is affected.


adria0:
>     * This is where the problem was.

Which one?

parity/informant.rs
```diff
@@ -224,7 +224,11 @@ impl<T: InformantData> Informant<T> {
 
     pub fn tick(&self) {
         let now = Instant::now();
-        let elapsed = now.duration_since(*self.last_tick.read());
```
rakita, line 227: *here. It happened that `last_tick` was lower than `elapsed`. and in this case `last_tick-elapes` can panic

adria0 (approved):

PR #208, opened 2020-12-29
Executable queue for ancient blocks inclusion 

rakita:
- Old ancient block queue will fill all requests to IoService and with that drain threads and stop the execution of other parts of the system. Here, inclusion is done one by one in its own thread.
- Additionally, there wes no flag that will stop the client from asking for new ancient blocks and we would ask for requests that will be abandon because of the full queue.  In this pull request, I pause download if 80% ancient queue is full, 80% is an arbitrary number and probably can be reduced because it does not represent how many requests we did send and expect to receive. For comparison with ordinary blocks, this is set to pause if the queue has anything in it even one item.


rakita:
execution tests as `cargo test --all` passes on Linux machine

rakita:
It is ready to be included. After reth-util is merged into `reth` I will switch link to that repo before merging this.

karim-agha (approved):

crates/ethcore/src/snapshot/tests/service.rs
```diff
@@ -363,6 +363,8 @@ fn recover_aborted_recovery() {
         }
         e => panic!("Snapshot restoration must be ongoing ; {:?}", e),
     }
+    // abort restoration so that we can delete snapshot root folder
+    service.abort_restore();
```
jomuel, line 367: why call it twice?

jomuel (approved):

crates/ethcore/src/snapshot/tests/service.rs
rakita: Poor attempt to test something. Will remove it.

PR #237, opened 2021-01-21
Local EN/DE block tests

rakita:


crates/ethcore/types/src/transaction/transaction_id.rs
```diff
@@ -23,7 +23,17 @@ use std::convert::TryFrom;
 #[repr(u8)]
 pub enum TypedTxId {
     AccessList = 0x01,
-    Legacy = 0x80, // With 0x80 we are sure that all other types will not overlap
```
rakita, line 26: This change is not relates to testing but Legacy Tx are represented as 0x00.

crates/ethcore/src/json_tests/local.rs
```diff
@@ -0,0 +1,203 @@
...
+        let block = match block {
+            Ok(block) => block,
+            Err(decoder_err) => {
+                warn!(target: "json-tests", "Error decoding test block: {:?}", decoder_err);
+                failed.push(name.clone());
+                continue;
+            }
+        };
+        if !is_same_block(&ref_block, &block) {
+            println!("block failed {:?}", block);
```
adria0, line 35: `err!`?

crates/ethcore/src/json_tests/local.rs
```diff
@@ -0,0 +1,203 @@
...
+    rlps.append(&block.header);
+    UnverifiedTransaction::rlp_append_list(&mut rlps, &block.transactions);
+    rlps.append_list(&block.uncles);
+    rlps.out()
+}
+
+pub fn is_same_block(ref_block: &Block, block: &Unverified) -> bool {
+    let test_exp = |exp: bool, err: &str| -> bool {
+        if !exp {
+            println!("Test mismatch on:{}", err);
```
adria0, line 56: `err!` ? (or `warn!`...)

adria0: I think that also is good to display which test failed additionally of the error message. Tests are executed in parallel and do you do not have the reference of which test this error refers to.

crates/ethcore/src/json_tests/local.rs
```diff
@@ -0,0 +1,203 @@
...
+                    continue;
+                }
+            } else {
+                TypedTxId::Legacy
+            };
+            is_ok = is_ok
+            && test_exp(tx.tx().nonce == ref_tx.nonce.0 , "Tx nonce")
+            && test_exp(tx.tx().gas_price == ref_tx.gas_price.0, "Tx gas price")
+            && test_exp(tx.tx().gas == ref_tx.gas_limit.0, "Tx gas")
+            //&& test_exp(tx.tx().action == ref_tx.)
```
adria0, line 137: Remove commented code?

adria0 (approved):
LGTM with comments

crates/ethcore/src/json_tests/local.rs
rakita: Test runner uses println for printing error

crates/ethcore/src/json_tests/local.rs
rakita: Failed test names are printed after all JSON test finish execution and this output will get clogged with them, I left it because it is useful when you are testing a subset of use cases or one specific tests.

PR #243, opened 2021-02-01
Yolo3 maximumExtraDataSize

rakita:
Clique needs more extradata than ethash. Config value is from Goerli.


adria0:
Superseeded by https://github.com/openethereum/openethereum/pull/244

PR #245, opened 2021-02-01
Tx view eip2930

rakita:
It seems only used for `gas_price_corpus` RPC


crates/ethcore/types/src/views/typed_transaction.rs
```diff
@@ -0,0 +1,246 @@
...
+
+/// View onto transaction rlp. Assumption is this is part of block.
+/// Typed Transaction View. It handles raw bytes to search for particular field.
+/// Access tx:
+/// 1 | [chainId, nonce, gasPrice, gasLimit, to, value, data, access_list, senderV, senderR, senderS]
+/// Legacy tx:
+/// [nonce, gasPrice, gasLimit, to, value, data, senderV, senderR, senderS]
+pub struct TypedTransactionView<'a> {
+    rlp: ViewRlp<'a>,
+}
```
adria0, line 35:
```suggestion
pub struct TypedTransactionView<'a> {
    rlp: ViewRlp<'a>,
    type: TypedTxId,
}
```

If we want to call the mehods `nonce`, `gas_price`, etc... the `transaction_type()` will be called multiple times with some computation.

Why not to get the transaction type and set into the `TypedTransactionView::new(...)` ?

rakita: It makes sense, will change it.

rakita:
In this pull, I enabled RPC calls with EIP-2930 transactions and renamed fields from `tx_type` to `transaction_type`.

crates/ethcore/types/src/views/typed_transaction.rs
```diff
@@ -51,11 +53,15 @@ impl<'a> TypedTransactionView<'a> {
...
+        TypedTransactionView {
+            rlp: rlp,
+            transaction_type,
+        }
     }
 
-    pub fn transaction_type(&self) -> TypedTxId {
-        let ref rlp = self.rlp.rlp;
+    /// Extract transaction type from rlp bytes.
+    fn extract_transaction_type(rlp: &Rlp) -> TypedTxId {
```
adria0, line 64: I would prefer:

```
let tx = rlp.data().unwrap();
let id = tx[0].try_into().unwrap();
```

to 

```
let tx = rlp.data().expect("unable to decode tx rlp");
let id = tx[0].try_into().expect("unable to decode tx type");
```

better for me is, returning `Option<TypedTxId>`

adria0 (approved):
LGTM with comments

crates/ethcore/types/src/views/typed_transaction.rs
rakita: Yep, it is better. Change is made.

PR #251, opened 2021-02-03
Remove pending secretstore

adria0:
Secret store is no longer supported, this PR removes all secret store features already existing on the client


selemis-organization:
Hi @adria0 / @rakita, perhaps I'm missing something but is it still the plan to remove the secretstore? Why was this PR closed?

PR #257, opened 2021-02-05
Give IoService a consistent interface

jomuel:
The start() method had two different signatures depending on configuration. Since ethcore is built with either config depending on the build target, this resulted in compiler errors for either evmbin or oe.


rakita (approved):
lgtm

adria0 (approved):

PR #259, opened 2021-02-07
Add back in openethereum.service

BriungRi:
This service file is mentioned in the README.md and in the docs. I wonder if it was accidentally deleted or the docs need to get updated.

Thanks


rakita:
In `main` it is named `parity.service` but in `dev` it is `openethereum.service`

PR #261, opened 2021-02-08
EIP-2930 spec change for signature hash

rakita:
Change on how EIP-2930 transaction is calculating a hash for signature


karim-agha (approved):

jomuel (approved):

adria0 (approved):
LGTM

PR #268, opened 2021-02-13
Sunsetting pWASM - part 1 of N

karim-agha:
This is part of a larger set of changes that aim to remove completely all WASM-VM related code.

In this change I'v created a tool that downloads the entire Kovan blockchain history from Infura. Otherwise sync takes forever and never completes even after few days.

The downloaded blocks are going to be used to reconstruct the statediff objects
produced by WASM contracts. Those statediffs will be hardcoded in a json file and
used in place of any historical WASM contract during block validation and inclusion.

Next step is to process all those blocks, and for each transaction that calls a wasm contract, record the execution and store the state diff to a file, then create one big JSON migration file that has all the known wasm transactions and their diffs.

### DIFF
--- a/bin/ethdl/src/main.rs
+++ b/bin/ethdl/src/main.rs
@@ -103,8 +103,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
     
   let pb = ProgressBar::new(rangelen);
   pb.set_style(ProgressStyle::default_bar()
-    .template("{spinner:.green} {percent}% [{elapsed_precise}] [{wide_bar:.cyan/blue}] {pos}/{len} - {per_sec} blocks/s - ETA {eta_precise}")
-    .progress_chars("#>-"));
+    .template(&format!("{}{}",
+    "{spinner:.green} {percent}% [{elapsed_precise}] ", 
+    "[{wide_bar:.cyan/blue}] {pos}/{len} - {per_sec} - ETA {eta}")));
 
   while let Ok(_) = blocks_stream.next().await.unwrap() {
     pb.inc(1);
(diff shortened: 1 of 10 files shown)


karim-agha:
*note: will fix spacing and formatting just noticed that my local machine has 2-space indents and the code 4-space indents.

bin/ethdl/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+  match Path::new(&filename).exists() {
+    true => Ok(false),
+    false => {
+      serde_json::to_writer_pretty(
+        &std::fs::File::create(&filename).unwrap(), 
+        &reqwest::Client::new()
+          .post(&request_url(&options))
+          .json(&serde_json::json!({
+            "jsonrpc": "2.0",
+            "id": SystemTime::now().elapsed().unwrap().subsec_nanos(),
```
jomuel, line 76: This seems to be an awkward way to generate a random id, `.elapsed().unwrap()` might even crash. Why not use a library PRNG? (It's not that important for a small tool like this though)

jomuel (approved):

bin/ethdl/src/main.rs
adria0: AFAIK, this id is only used by the client, the server just copies this value in the response. So, in this case this is not especially relevant.

karim-agha: @jomuel That's right, I was just looking for a quick one-liner to use here without having to have a global static random number generator object. For "real" code this is obviously not a good idea what I'm doing here.

@adria0 I wasn't aware of the implementation details and just used a random number in case the server tries to correlate responses with requests using this field.

In any case, with Dragan's dump that I'm downloading right now, this code will not be used very much anymore.

Issue #89, opened 2020-11-10
Performance: v3.1.0 vs v3.0.1


mdben1247:
- **OpenEthereum version**: 3.1.0
- **Operating system**: Linux
- **Installation**: built from source
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

We are comparing two nodes, one 3.0.1 and another v3.1.0. Both are running on the same machine with the same configuration parameters (fat DB, lots of cache, total about 30GB) and have been built from source with the same compiler and settings. 3.1.0 db was converted from 2.5.13 db using the provided tool. 

The workload is continuous serving of websocket RPC requests in parallel for 5-10 clients - per block subscriptions, pending transaction subscriptions and state queries. When measuring performance, the newly released 3.1.0 fares much worse. Block import durations as reported in openethereum log, are about 2-3 times slower. Client measured durations to complete all relevant WS requests after receiving a new block are also about 2-3 times slower. 

I realize 3.1.0 is based on 2.5.13, but are there any obvious differences with regards to 3.0.1 that may account for such a big performance hit?

adria0:
@mdben1247, we backported from 2.5.13 because 3.0.1 was really unstable in some environments, and we priorized stability first. Maybe one of the performance issues is due that 2.5.13 uses some old libraries, and probablly updating them (the same that used 3.0.1) could improve the performance. Currently we plan to update libraries by 1Q2021, after BerlinHF.

If you could provide the performance measuraments you did (including which calls are you doing) it will provide to us an unvaluable information to analize it.

mdben1247:
@adria0 I agree, it could the libraries. For quick comparison, I am attaching logs from both versions running in parallel, cold started at the same time with same configurations. Its pretty clear block import times are about 2-3x larger in the new version. That should be easy to replicate. As for measuring other calls, while I cannot copy our production environment, I am currently writing a test benchmark that captures its main features. I'll share it when it's done.

On a side note, I realize 3.0.1 is not supported anymore, but if you'd make an educated guess, was it pretty much Berlin-ready or were some major features missing?

[openethereum_301.txt](https://github.com/openethereum/openethereum/files/5523951/openethereum_301.txt)
[openethereum_310.txt](https://github.com/openethereum/openethereum/files/5523953/openethereum_310.txt)

adria0:
[...]

PR #270, opened 2021-02-15
kvdb-* upgrade

mdben1247:
Builds on #253. Upgraded `kvdb-*`, `memory-db`, `trie-db`, `hash-db`, `parking_lot` libraries.


adria0:
> Builds on #253. Upgraded `kvdb-*`, `memory-db`, `trie-db`, `hash-db`, `parking_lot` libraries.

@mdben1247, we had in the past a lot of performance regression problems updating the kvdb libraries. This change requires a full performance diff analysis, especially on disk usage.

mdben1247:
As reported here #89, our own testing shows about 2x faster block syncing / verification times and execution for various rpc workloads. That's quite a lot actually. But of course, more testing is needed here, I'll be doing just that and report back.

adria0:
> As reported here #89, our own testing shows about 2x faster block syncing / verification times and execution for various rpc workloads. That's quite a lot actually. But of course, more testing is needed here, I'll be doing just that and report back.

Please, check what happens when openetherum 3.0.0 used an updated version of the DB in https://etherscan.io/chartsync/chaindefault between may 26th and 14th jun.

mdben1247:
> Please, check what happens when openetherum 3.0.0 used an updated version of the DB in https://etherscan.io/chartsync/chaindefault between may 26th and 14th jun.

I see what you mean. I've been mostly focused on speed for now. I'll take a better look at disk usage.

mdben1247:
About disk usage. Newer `rocksdb` seem to use write ahead logs more extensively.

https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log

By default there is some heuristics which determines the maximum total size of WAL and in my test case equals to a about 50GB (ie. a lot). The db size is then some 'normal' size plus WAL. At client startup the db will be cleaned up, so it starts with 'normal' size and zero WAL. As the client syncs WAL will grow and occasionally they are partially deleted, so the whole db size will be 'normal' size plus some variable amount of WAL. This seem to be the disk usage spikes that we see. If one doesn't care much about disk usage, this is all probably just fine. But if we do care, it would make sense to limit the size of WAL.

There is an option for that in `rocksdb`: [max_total_wal_size](https://github.com/facebook/rocksdb/blob/48bfca38f6f175435052a59791922a1a453d9609/include/rocksdb/options.h#L467). Unfortunately it is not exposed through `kvdb-rocksdb`. For testing purposes I merged a patched `kvdb-rocksdb 0.9.1` into the tree and added a command line option to the client to expose the setting `--db-max-wal-size`. My testing shows this limit is respected, WAL never exceed the max. While syncing 4-5 blocks/second, WAL reach 10GB after 15 minutes. So I am thinking a rather small max value (perhaps 4 GB) would be appropriate for normal operations. 

Of course this all needs a more testing. Rushing here seems unwise, however I think large performance gains of the newer `rocksdb` make this worthy of serious consideration.

mdben1247:
PR to add max wal limit to `kvdb-rocksdb`: https://github.com/paritytech/parity-common/pull/528

ytrezq:
@mdben1247 is this changed backward compatible? Or does if I test this pull request code, I will no longer be able to use the database with previous versions of OpenEthereum because of format changes?

ytrezq:
I’m seeing it’s far to to be an up to date rocksdb version which is to be used. Recent versions added support for avx2 and the new `io_uring` low filesystem latency system call family.

mdben1247:
> is this changed backward compatible? Or does if I test this pull request code, I will no longer be able to use the database with previous versions of OpenEthereum because of format changes?

I don't know, I haven't tried it myself. I'd strongly suggest backing up first.

> I’m seeing it’s far to to be an up to date rocksdb version which is to be used. Recent versions added support for avx2 and the new `io_uring` ultra low latency filesystem access system call family.

`openethereum` is using `kvdb-rocksdb` which doesn't use the latest `rocksdb` version. Long term I agree it would make sense to keep up with new versions, but there may be quite a lot of work involved.

rakita:
Closing pull request in favor of: https://github.com/openethereum/openethereum/pull/332

PR #271, opened 2021-02-17
Change ProtocolId to U64

rakita:
* Change ProtocolId to support variable subprotocol names
* Add yolo3x testnet


bin/oe/run.rs
```diff
@@ -227,12 +227,10 @@ pub fn execute(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<RunningClient
         Some(id) => id,
         None => spec.network_id(),
     };
-    if spec.subprotocol_name().len() != 3 {
-        warn!("Your chain specification's subprotocol length is not 3. Ignoring.");
+    if spec.subprotocol_name().len() > 8 {
```
karim-agha, line 230: Nitpicking, might not be relevant or too important, but do we want also to check if the protocol name is not empty?

karim-agha: Also more netpicking and most likely even less relevant, but the protocol name is a unicode string, which in most sane cases will contain only ascii chars and makes this check valid, however "sńap" is not of length 4, because it has accent letters. I would probably compare the length of the raw bytes instead of string chars in this case.

karim-agha (approved):
LGTM

bin/oe/run.rs
karim-agha: I was wrong. Actually, `String::len()` returns the number of bytes. Just looked up the docs:

```
/// Returns the length of this `String`, in bytes, not [`char`]s or
/// graphemes. In other words, it may not be what a human considers the
/// length of the string.
```

jomuel (approved):
lgtm

bin/oe/run.rs
rakita: Nitpick away :)
Not the prettiest part but it seems okay. If it is empty I assume that U64 will be zero.

PR #284, opened 2021-02-26
Sunce86/rpc module reverted for RPC JSON api

sunce86:


bin/oe/cli/mod.rs
```diff
@@ -409,9 +409,9 @@ usage! {
...
+            ARG arg_jsonrpc_apis: (String) = "web3,eth,pubsub,net,parity,parity_pubsub,traces,rpc", or |c: &Config| c.rpc.as_ref()?.apis.as_ref().map(|vec| vec.join(",")),
             "--jsonrpc-apis=[APIS]",
-            "Specify the APIs available through the HTTP JSON-RPC interface using a comma-delimited list of API names. Possible names are: all, safe, debug, web3, net, eth, pubsub, personal, signer, parity, parity_pubsub, parity_accounts, parity_set, traces, secretstore. You can also disable a specific API by putting '-' in the front, example: all,-personal. 'safe' enables the following APIs: web3, net, eth, pubsub, parity, parity_pubsub, traces",
+            "Specify the APIs available through the HTTP JSON-RPC interface using a comma-delimited list of API names. Possible names are: all, safe, debug, web3, net, eth, pubsub, personal, signer, parity, parity_pubsub, parity_accounts, parity_set, traces, rpc, secretstore. You can also disable a specific API by putting '-' in the front, example: all,-personal. 'safe' enables the following APIs: web3, net, eth, pubsub, parity, parity_pubsub, traces, rpc",
```
karim-agha, line 414: Just a small suggestion: Perhaps we would like to have the list of all enabled APIs string generated automatically, so that in the future we will not have to remember changing this list in all those places?

bin/oe/rpc_apis.rs
```diff
@@ -146,6 +154,30 @@ impl FromStr for ApiSet {
     }
 }
 
+fn to_modules(apis: &HashSet<Api>) -> BTreeMap<String, String> {
```
karim-agha, line 157: neat! :)

karim-agha (approved):
Neat! LGTM.

rakita (approved):
lgtm

bin/oe/cli/mod.rs
sunce86: Yeah, nice suggestion. But for this particular situation it might not be worth it, since lists are not exactly the same for http,ws and ipc. Therefore we would need to introduce one string generator to replace only two strings.

Issue #277, opened 2021-02-24
Block sync stopped without any errors.


rlaace423:
For questions please use https://discord.io/openethereum, issues are for bugs and feature requests.

_Before filing a new issue, please **provide the following information**._

- **OpenEthereum version (>=3.1.0)**: 3.1.1
- **Operating system**: Linux
- **Installation**: pre-built binary
- **Fully synchronized**: yes (warp=false, tracing="on", cache_size=4096)
- **Network**: ethereum
- **Restarted**: I'm trying to stop..

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

Openethereum stopped sync without any errors in log about 4 hours before(2021-02-23T20:26:17). Process itself is still alive, but it doesn't sync new blocks since 11915524.
![image](https://user-images.githubusercontent.com/5516480/108934037-847b3500-768f-11eb-8996-edcdcdf48ff1.png)

As you can see in the above, It shows any errors. Now I'm trying to stop it, but It hangs now.
Also there's enough disk space **(total 1.2TiB, 51% used)**, ram usage **(total 1.2GiB, 7.34GiB used currently).**

When It stopped, other system indicators seem to be normal, See below.

![image](https://user-images.githubusercontent.com/5516480/108933786-1afb2680-768f-11eb-90e0-54748ee03dc8.png)
cpu usage

![image](https://user-images.githubusercontent.com/5516480/108933931-5564c380-768f-11eb-927c-12ee4a7f247c.png)
current disk usage

![image](https://user-images.githubusercontent.com/5516480/108933198-14b87a80-768e-11eb-9817-a8a3d62ddbf7.png)
Network Input(byte), Network Output(byte), Network Packet Input(count), Network Packet Output(count) in order.

![image](https://user-images.githubusercontent.com/5516480/108934370-2a2ea400-7690-11eb-8a0d-e1053ff0f5fb.png)
remaining RAM

How can I solve this? Is it okay that I kill openethereum process forcibly? Does it cause database break down?
Is there anyway to stop and start openethereum gracefully?

Legogris:
I had this on 3.0.1 a handful of times - a process restart resolved it for me.

It's a bit troubling to say the least, though.

ytrezq:
@rlaace423 just send `SIGINT` to the process for a gracefull shutdown.

yorickdowne:
[...]

PR #286, opened 2021-03-01
Backport: Block sync stopped without any errors. #277

rakita:
Full discussion here: https://github.com/openethereum/parity-ethereum/issues/11758
again reported here in v3.1.1 : https://github.com/openethereum/openethereum/issues/277

This is only known instance where even Log Informant stopped reporting, and the reason why I assume these are same issues.


jomuel (approved):

adria0 (approved):
Replaces `parking_lot::RwLock` with `std::sync::StdRwLock`
Adds a ` priority_tasks_gate` for executing only one priority timer at once

LGTM

PR #287, opened 2021-03-01
Send RLPx auth in EIP-8 format

rakita:
Sending part is taken from this pull request:https://github.com/openethereum/openethereum/pull/81
but omitting part of removing inbound support for the old protocol, full removal should be done London hardfork.


jomuel (approved):
lgtm. Also much more readable with the use of `RlpStream`

sunce86 (approved):
Looks good.

PR #289, opened 2021-03-02
ethcore/snapshot: fix double-lock in Service::feed_chunk

BurtonQin:
A double-lock bug can happen in `Service::feed_chunk()`:
The first lock is at L763, the second at L769.
The fix reuses the `restoration` returned by the first lock.
https://github.com/openethereum/openethereum/blob/efb80e10324973223a8d7e64941b795aa6156e48/crates/ethcore/src/snapshot/service.rs#L761-L769


rakita (approved):
Thank you

karim-agha (approved):

PR #290, opened 2021-03-02
AuRa multi block reward

rakita:
Requested in discord.
It does not change old specs, it adds support for changing block rewards in AuRa spec.
Same code is used in ethash spec.

### DIFF
--- a/crates/ethcore/src/engines/authority_round/mod.rs
+++ b/crates/ethcore/src/engines/authority_round/mod.rs
@@ -80,7 +80,7 @@ pub struct AuthorityRoundParams {
     /// Immediate transitions.
     pub immediate_transitions: bool,
     /// Block reward in base units.
-    pub block_reward: U256,
+    pub block_reward: BTreeMap<BlockNumber, U256>,
     /// Block reward contract transition block.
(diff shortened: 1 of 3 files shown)


sunce86 (approved):
Looks good.

crates/ethcore/src/engines/authority_round/mod.rs
```diff
@@ -113,7 +113,24 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
+                    ret.insert(0, U256::zero());
+                    ret
+                },
+                |reward| match reward {
+                    ethjson::spec::BlockReward::Single(reward) => {
+                        let mut ret = BTreeMap::new();
+                        ret.insert(0, reward.into());
+                        ret
+                    }
+                    ethjson::spec::BlockReward::Multi(multi) => multi
```
adria0, line 128: I think that we have to check here that there is the block zero defined, instead of crashing later with `Current block's reward is not found; this indicates a chain config error; qed`. Ideally implement `try_from` instead `from`

adria0 (approved):
LGTM with comments

PR #291, opened 2021-03-02
Migrate compare_and_swap to compare_exchange

gakada:
Currently, building generates warnings about `compare_and_swap` being deprecated, so this [migrates](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#migrating-to-compare_exchange-and-compare_exchange_weak) it to `compare_exchange`.

### DIFF
--- a/bin/oe/run.rs
+++ b/bin/oe/run.rs
@@ -450,7 +450,15 @@ pub fn execute(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<RunningClient
     let is_ready = Arc::new(atomic::AtomicBool::new(true));
     miner.add_transactions_listener(Box::new(move |_hashes| {
         // we want to have only one PendingTransactions task in the queue.
-        if is_ready.compare_and_swap(true, false, atomic::Ordering::SeqCst) {
+        if is_ready
+            .compare_exchange(
+                true,
+                false,
+                atomic::Ordering::SeqCst,
+                atomic::Ordering::SeqCst,
+            )
+            .is_ok()
+        {
             let task =
                 ::sync::PriorityTask::PropagateTransactions(Instant::now(), is_ready.clone());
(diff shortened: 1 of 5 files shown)


adria0 (approved):
LGTM, it follows the migration rules specified in https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#migrating-to-compare_exchange-and-compare_exchange_weak

bin/oe/run.rs
```diff
@@ -450,7 +450,13 @@ pub fn execute(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<RunningClient
     let is_ready = Arc::new(atomic::AtomicBool::new(true));
     miner.add_transactions_listener(Box::new(move |_hashes| {
         // we want to have only one PendingTransactions task in the queue.
-        if is_ready.compare_and_swap(true, false, atomic::Ordering::SeqCst) {
+        if is_ready.compare_exchange(
+            true,
+            false,
+            atomic::Ordering::SeqCst,
+            atomic::Ordering::SeqCst,
+        ) == Ok(true)
```
rakita, line 458: I little bit dislike that we have `true` in `Ok(true)` it doesn't feel nice to have it in two places, especially if it could be changed to become:
`if let Ok(_) = is_ready.compare_exhange(....) {`
Can you change all others in same way?

adria0: I missed this.

What about  ` if is_ready.compare_exhange(....).is_ok()` ?

rakita: even better :D

gakada: Switched to `is_ok`/`is_err`, is it also ok to remove a temporary `prev_last_sealed_block` (which I did)?

rakita (approved):
lgtm

PR #292, opened 2021-03-02
Disable EIP-2315

rakita:
Still need to disable it from berlin_test chainspec, but am waiting for tests to be upgraded before that.


rakita:
Leaving it enabled for berlin hardfork

PR #295, opened 2021-03-04
Implement the filter argument in parity_pendingTransactions

### LINKED ISSUE #159: Can no longer apply filters to parity_pendingtransactions
This is a regression of https://github.com/openethereum/parity-ethereum/issues/11189. Using the request of the doc, OpenEthereum complains that only one parameter is accepted not 2.

jomuel:
Closes #159


rakita:
this is command for fmt: `cargo fmt --all -- --config merge_imports=true`

crates/rpc/src/v1/impls/parity.rs
```diff
@@ -268,16 +268,28 @@ where
...
-    fn pending_transactions(&self, limit: Option<usize>) -> Result<Vec<Transaction>> {
+    fn pending_transactions(
+        &self,
+        limit: Option<usize>,
+        filter: Option<TransactionFilter>,
+    ) -> Result<Vec<Transaction>> {
         let ready_transactions = self.miner.ready_transactions(
             &*self.client,
-            limit.unwrap_or_else(usize::max_value),
+            usize::max_value(),
```
rakita, line 278: you omitted `limit` here?

rakita: This will always copy every pending tx from pool and then do the filtering below.

crates/rpc/src/v1/types/transaction_filter.rs
```diff
@@ -0,0 +1,75 @@
+use ethereum_types::{Address, U256};
+use v1::types::Transaction;
+
+#[allow(non_camel_case_types)]
+#[derive(Debug, Deserialize, Serialize)]
+#[serde()]
+pub enum EqFilterArgument<T: Eq> {
+    eq(T),
+    Nil,
```
rakita, line 9: `None` is a more used noun than `Nil` in rust.

crates/rpc/src/v1/types/transaction_filter.rs
```diff
@@ -0,0 +1,75 @@
...
+        match self {
+            ValueFilterArgument::eq(expected) => value == expected,
+            ValueFilterArgument::lt(threshold) => value < threshold,
+            ValueFilterArgument::gt(threshold) => value > threshold,
+            ValueFilterArgument::Nil => true,
+        }
+    }
+}
+
+#[allow(non_camel_case_types)]
```
rakita, line 54: camel case is used for json rpc

rakita: I see that original pull request used `gas_price` but all other jsonrpc filed are camel case and in this case I think it is better to be consistent and use `gasPrice`

rakita (changes requested):
See other comment, and filter is missing `action: Value - Set to contract_creation for contract creation transactions.`

jomuel:
Two remarks:

- Now that the filtering has been moved, the test I wrote doesn't fully cover the changes anymore (using `TestMinerService` instead of the changed `Miner`). I didn't change it as it still covers enough - request parsing and filtering.
- Inside the `ready_transactions_filtered` function, two different closures are used depending on settings (`from_queue` and `from_pending`). The original PR in parity-ethereum only added filtering to `from_pending`. As I could find no reason for this, I also added it to `from_queue`. For the no-filter case the old implementation with cached transaction lists is still used.

crates/concensus/miner/Cargo.toml
```diff
@@ -30,7 +30,10 @@ log = "0.4"
 parity-runtime = { path = "../../runtime/runtime" }
 parking_lot = "0.7"
 price-info = { path = "./price-info", optional = true }
-rlp = { version = "0.3.0", features = ["ethereum"] }
+rlp = { version = "0.4.6" }
+serde = "1.0"
+serde_derive = "1.0"
```
karim-agha, line 35: Afaik newer versions of serde have serde_derive built in and don’t need a separate import.

rakita: `serde = { version = "1.0", features = ["derive"] }` this should probably work. https://serde.rs/derive.html

rakita (approved):
lgtm. See to change import and if no problems merge after

crates/concensus/miner/Cargo.toml
jomuel: Yes, that worked.

karim-agha (approved):

pedrobranco:
Any expected release date which includes this fix?

PR #308, opened 2021-03-09
Fixing a compilation warning

karim-agha:


crates/ethcore/sync/src/chain/mod.rs
```diff
@@ -489,7 +489,8 @@ impl ChainSyncApi {
 
         if self
             .priority_tasks_gate
-            .compare_and_swap(false, true, Ordering::AcqRel)
+            .compare_exchange(false, true, Ordering::Acquire, Ordering::Release)
+            .is_ok()
```
rakita, line 493: I think `is_err()` is needed here. If `compare_exchange` does not set new value it returns err, which means that the current value is true and somebody has already taken access to gate.

karim-agha: Right!... fixing it.

rakita (approved):
lgtm

PR #310, opened 2021-03-09
Gradually upgrade crates to Rust 2018 edition

karim-agha:
This changeset upgrades `common-types` and `ethjson` to 2018 edition.

### DIFF
--- a/scripts/docker/alpine/Dockerfile
+++ b/scripts/docker/alpine/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.12.3 AS builder
+FROM alpine:3.13.2 AS builder
 
 # show backtraces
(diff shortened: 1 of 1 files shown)


jomuel (approved):
lgtm

karim-agha:
Don't merge this yet please, need to apply the fix from #312

rakita:
Fix is already applied and merged in `dev`
for ethjson, change is already done in master, I will see to push `merge` commit to dev so that we can connect `main` and `dev` branched properly, this would make it easier in future to merge `dev` to `main` when we want to make new release.

adria0:
`Submodule json_tests updated 6031 files` 

It's upgrading also the json_tests, I think that should be removed from this PR

Issue #249, opened 2021-02-03
ethereum-types upgrade guidance


mdben1247:
Basing on `v3.1.1`, I upgraded `ethereum-types` to `0.9.2` (with a host of other stuff that was pulled along). When running tests (excluding `wasm` and `secret-store`) I get 4 failures for POA stuff that I can't figure out what could be wrong, everything else is OK though. But more troubling, when running this version live, some blocks (every 50 or so) would give InvalidStateRoot. I pretty much double checked everything. So, if anyone has any clues on where to look and tips for debugging this I'd appreciate it.

rakita:
It is hard to tell devil is always in the detail. can you share what exactly failures are for POA it is maybe related. And you can make draft pull request for me to see changes, there are probably a lot of them. And can you move changes to be based on `dev` branch, this is the branch that we will use to create next release.

mdben1247:
POA failures are
```
---- engines::authority_round::tests::rejects_future_block stdout ----
thread 'engines::authority_round::tests::rejects_future_block' panicked at 'assertion failed: engine.verify_block_external(&header).is_ok()', ethcore/src/engines/authority_round/mod.rs:2015:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- engines::authority_round::tests::proposer_switching stdout ----
thread 'engines::authority_round::tests::proposer_switching' panicked at 'assertion failed: engine.verify_block_external(&header).is_ok()', ethcore/src/engines/authority_round/mod.rs:1989:9

---- snapshot::tests::proof_of_authority::fixed_to_contract_only stdout ----
thread 'snapshot::tests::proof_of_authority::fixed_to_contract_only' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Snapshot(BadEpochProof(2)), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })', ethcore/src/snapshot/tests/proof_of_authority.rs:259:87

---- snapshot::tests::proof_of_authority::fixed_to_contract_to_contract stdout ----
thread 'snapshot::tests::proof_of_authority::fixed_to_contract_to_contract' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Snapshot(BadEpochProof(2)), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })', ethcore/src/snapshot/tests/proof_of_authority.rs:302:87

```
[...]

PR #315, opened 2021-03-11
mbdben ethereum-types upgrade 

rakita:
#249
Upgraded ethereum-types and other libs:

ethereum-types -> 0.9.2
rlp -> 0.4.6
keccak-hash -> 0.5.0
parity-crypto -> 0.6.2
ethabi -> 12.0.0
ethabi-derive -> 12.0.0
ethabi-contract -> 11.0.0
ethbloom -> 0.9.1
rand -> 0.7.3
trie-standardmap -> 0.15.2
triehash -> 0.5.0


karim-agha (approved):

crates/ethcore/src/miner/stratum.rs
```diff
@@ -47,6 +47,14 @@ pub struct Options {
     pub secret: Option<H256>,
 }
 
+fn clean_0x(s: &str) -> &str {
```
jomuel, line 50: This function is added twice in the PR, here and in bin/oe/helpers, could it be moved to utils?

jomuel (approved):
Why are we replacing standard `HeapSizeOf` with parity `MallocSizeOf`? `HeapSizeOf` should be derive-able too?

rakita:
> Why are we replacing standard `HeapSizeOf` with parity `MallocSizeOf`? `HeapSizeOf` should be derive-able too?
Servo stopped maintaining HeapSizeOf and we are using their internal lib MallocSizeOf: https://github.com/servo/heapsize

crates/ethcore/src/miner/stratum.rs
rakita: It is a one `if` function, and I am fine with it for now.

PR #316, opened 2021-03-12
Updating ethcore and ethjson crates to Rust Edition 2018

karim-agha:
This change updates the rust edition for ethcore and ethjson to Rust 2018.


crates/ethcore/res/chainspec/test/yolo3_test.json
```diff
@@ -0,0 +1,204 @@
+{
```
adria0, line 1: IMO this should be added in another PR

karim-agha: Agreed.

jomuel (approved):

crates/ethjson/src/state/log.rs
```diff
@@ -50,3 +53,13 @@ mod tests {
         // TODO: validate all fields
     }
 }
+
+impl From<Log> for LogEntry {
```
rakita, line 57: Is this even used, do you need it else somewhere else?

crates/ethjson/src/transaction/transaction.rs
```diff
@@ -45,10 +47,37 @@ pub struct Transaction {
     pub v: Uint,
 }
 
+impl From<Transaction> for UnverifiedTransaction {
```
rakita, line 50: I think this is removed from code because it is unused. Do you have usage for it?

rakita (approved):
lgtm. Just you can remove two added `From` `Transaction` and `Log` and can add them in pull request when you need them. name of pull request could be without `update ethcore/ethjson`. And btw great job I wanted to do this for months :)

crates/ethjson/src/transaction/transaction.rs
karim-agha: Afaik it’s used in tests. It’s also part of removing the dependency of ethcore on ethjson. I’m having lunch now and I’ll double check it ASAP when I get home.

rakita: It was modified by dusan and removed when test started supporting AccessList, it is deadcode now.

karim-agha: Sure, I'll remove it in a follow-up PR.

PR #317, opened 2021-03-12
Send SIGTERM instead of SIGHUP to OE daemon

eskp:
`SIGHUP` tells an application that shell/terminal has disappeared, forcing an application to react in unpredictable ways.
`SIGTERM` tells an application to terminate, which is exactly what we want here, giving it chance to complete and shut down cleanly.

Changing to `SIGTERM` we can see the following log when running `service openethereum restart`. Note the couple of lines OE finishing work:

```
Mar 12 06:27:21 ip-172-31-46-148 systemd[1]: Stopping OpenEthereum Daemon...
Mar 12 06:27:21 ip-172-31-46-148 openethereum[16321]: 2021-03-12 06:27:21 UTC Finishing work, please wait...
Mar 12 06:27:22 ip-172-31-46-148 openethereum[16321]: 2021-03-12 06:27:22 UTC StepService: engine is dropped; exiting.
Mar 12 06:27:22 ip-172-31-46-148 systemd[1]: Stopped OpenEthereum Daemon.
Mar 12 06:27:22 ip-172-31-46-148 systemd[1]: Started OpenEthereum Daemon.
```

With `SIGHUP` OE doesn't have a chance to exit cleanly and we don't see the finishing work log:

```
Mar 12 06:25:15 ip-172-31-46-148 systemd[1]: Stopping OpenEthereum Daemon...
Mar 12 06:25:16 ip-172-31-46-148 systemd[1]: Stopped OpenEthereum Daemon.
Mar 12 06:25:16 ip-172-31-46-148 systemd[1]: Started OpenEthereum Daemon.
```

Ref: https://stackoverflow.com/a/690631


scripts/openethereum.service
```diff
@@ -13,8 +13,10 @@ ExecStart=/usr/bin/openethereum --config /etc/openethereum/config.toml
 Restart=on-failure
 
 # Specifies which signal to use when killing a service. Defaults to SIGTERM.
-# SIGHUP gives openethereum time to exit cleanly before SIGKILL (default 90s)
-KillSignal=SIGHUP
+# SIGTERM gives openethereum time to exit cleanly (default 90s) before sending SIGKILL
+KillSignal=SIGTERM
+TimeoutStopSec=90
```
rakita, line 18: We have a registered handler for SIGTERM and I agree that it should be there. it is ctrlc lib: https://github.com/Detegr/rust-ctrlc
Looking at https://www.freedesktop.org/software/systemd/man/systemd.kill.html#, option `SendSIGKILL=yes` is default and could be removed. 
And can you put timeoutStopSec to something like 300 just to be sure.

eskp: Done

rakita (approved):
lgtm

karim-agha (approved):

PR #252, opened 2021-02-04
Remove eth/63 protocol version

rakita:
* Removes eth/63 version of the protocol

PR #333, opened 2021-03-22
Revert "Remove eth/63 protocol version (#252)"

rakita:
This reverts commit bbecb0415ef1ffe5d1c15d9a0fd19de859f0eb1a.

### DIFF
--- a/crates/ethcore/sync/src/api.rs
+++ b/crates/ethcore/sync/src/api.rs
@@ -32,7 +32,8 @@ use std::{
 
 use chain::{
     fork_filter::ForkFilterApi, ChainSyncApi, SyncState, SyncStatus as EthSyncStatus,
-    ETH_PROTOCOL_VERSION_64, PAR_PROTOCOL_VERSION_1, PAR_PROTOCOL_VERSION_2,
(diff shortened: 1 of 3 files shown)


jomuel (approved):
lgtm

adria0:
> This reverts commit [bbecb04](https://github.com/openethereum/openethereum/commit/bbecb0415ef1ffe5d1c15d9a0fd19de859f0eb1a).

Could you explain the reason for reverting the commit @rakita?

ngyam:
@adria0 we at energyweb with our AuRa chains kinda look forward to this, our nodes still stuck at 2.5.x and migration paths to 3.x seem to be broken for us. First we'd need to upgrade to a version that has 63/64 too with no showstopper bugs.
We need to have both until we migrate so there is a smooth transition and our chains not get rekt in the process :grimacing:
I guess this PR would ensure that.

rakita:
For kovan and energyweb it seems they don't have a big percentage of nodes that have support for `eth/`64 and they would need to have a transition from v2.5.13, it could be done with v3.1.1+(bugfix commit #318) but this seems like a simpler solution. 
@ngyam i would recommend you to cherry-pick this commit and use it on top of v3.2.1, v3.2.1 is tested and pushed for production.

adria0 (approved):
LGTM

PR #341, opened 2021-03-30
Refactoring ethcore::client::Client god-type, part 1 of N

karim-agha:
This change is the first in the series of refactoring the client object into a set of smaller managable pieces, the guidelines for those changes are:

  - Completely get rid if the `ethcore::client::traits` module and organize the traits defined there by their function rather than the type of language construct they are.
  - There are around 10+ different types called `BlockChain` encountered so far during this refactoring session. Reduce them to one.
  - Completely get rid of the `Client` struct and split its functionality into componentized types that interact with each other - preferably through message passing in an agent architecture.

In this first stab of splitting the `client.rs` file into smaller files I have made some internal member variables public so they are accessible from within other files. This will change with time, it's a temporary remedy.

*This change does not change any functional behavior of the code and the system should just work as it worked before.*

### DIFF
--- a/crates/ethcore/src/client/blockchain.rs
+++ b/crates/ethcore/src/client/blockchain.rs
@@ -21,8 +21,7 @@ use super::{
     info::{BlockChain, BlockInfo},
     io::IoClient,
     traits::TransactionRequest,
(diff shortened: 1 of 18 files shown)


jomuel (approved):

rakita (approved):
lgtm

Patriciamh (approved):

adria0:
👏

PR #352, opened 2021-04-15
Implement eth/65

jomuel:
Since eth/64 is about to be deprecated, we want to include the upgrade to eth/65 in the upcoming 3.2.2 release. We reuse @vorot93's implementation from the abandoned PR #38.

The review of #38 contains several suggestions (mainly for performance improvement), that could be included here or in a follow-up PR.


karim-agha (approved):
code-wise lgtm, does it sync correctly with those changes with mainnet?

jomuel:
> code-wise lgtm, does it sync correctly with those changes with mainnet?

I'll try to run it locally again to make sure. Do you have specific changes in mind that could break it? (To check the new messages are parsed/crafted correctly should be enough imo)

rakita (changes requested):
I will need to check this pull request a little bit more in detail. But these change is required for TypedTx EIP-2718.

sunce86 (changes requested):

crates/ethcore/sync/src/chain/mod.rs
```diff
@@ -156,6 +156,8 @@ impl From<DecoderError> for PacketProcessError {
     }
 }
 
+/// Version 65 of the Ethereum protocol and number of packet IDs reserved by the protocol (packet count).
+pub const ETH_PROTOCOL_VERSION_65: (u8, u8) = (65, 0x11);
```
rakita, line 160: The second number is the count of messages that we have for that protocol. Should be greater then eth/64

rakita (commented):
added few more comment

crates/ethcore/sync/src/chain/mod.rs
jomuel: Are you sure? This was already questioned in the old PR and review, but I think 0x11 is correct, as it is the count of reserved packet ids, not of actually used ones. (Neither protocol actually has 17 messages, but both use the range of 0x00 to 0x10, so 0x11. Afaict the number is not used anywhere.)

rakita: It is used here: https://github.com/openethereum/openethereum/blob/582bca385fedb1af682e989e5bcc6b3b2cf53028/crates/net/network-devp2p/src/session.rs#L458-L466
`packet_count` is that last number.

But you are right, I didn't see that message id's have a gap, and new packet id's are filled in that gap without increasing the last message id. Setting 0x11 is okay.

crates/ethcore/sync/src/chain/propagator.rs
```diff
@@ -238,15 +256,27 @@ impl SyncPropagator {
...
+                        } else {
+                            let appended = packet.append_raw_checked(
+                                &tx.encode(),
+                                1,
+                                MAX_TRANSACTION_PACKET_SIZE,
+                            );
+                            if !appended {
+                                // Maximal packet size reached just proceed with sending
+                                debug!(target: "sync", "Transaction packet size limit reached. Sending incomplete set of {}/{} transactions.", pushed, to_send.len());
+                                to_send = to_send.into_iter().take(pushed).collect();
```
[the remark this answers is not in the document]
jomuel: `to_send` should contain the transactions included in the packet, so I think `into_iter().take()` is correct. I added the same line in the `if is_hashes` block.

crates/ethcore/sync/src/chain/propagator.rs
```diff
@@ -238,15 +256,27 @@ impl SyncPropagator {
...
-                            break;
+                        if is_hashes {
+                            if pushed >= NEW_POOLED_HASHES_LIMIT {
+                                debug!(target: "sync", "NewPooledTransactionHashes length limit reached. Sending incomplete list of {}/{} transactions.", pushed, to_send.len());
+                                break;
+                            }
+                            packet.append(&hash);
+                            pushed += 1
+                        } else {
+                            let appended = packet.append_raw_checked(
```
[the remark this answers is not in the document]
jomuel: Done with the merging of the dev branch.

crates/ethcore/sync/src/chain/propagator.rs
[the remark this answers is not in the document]
rakita: The problem that I see is not if it okay to take N amount of transactions, but what exactly transactions are taken. `into_iter()`  visits all elements in **arbitrary order**, that would mean that we `take` N random transactions and not the transactions that we added to rlp. We probably need an additional `HashSet` of `send_tx` that we will return if there is more than MAX_TRANSACTION_PACKET_SIZE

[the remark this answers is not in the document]
jomuel: Yes, you are right. I will fix it.

jomuel:
I have been running a node with this eth-65 implementation over the weekend to test it and everything looks ok. This should be rtm.

crates/ethcore/sync/src/chain/propagator.rs
```diff
@@ -230,42 +248,53 @@ impl SyncPropagator {
             }
 
             // Construct RLP
-            let (packet, to_send) = {
```
rakita, line 233: here, we overwrote `to_send` with transactions that we intend to send. and we use it again when we are setting `peer_info.last_sent_transactions`. In the new commit, we will always set all `to_send` not just transactions that we intend to send to peer.

rakita: `stats` was not main problem, but setting `peer_info.last_sent_transactions` wrongly

rakita (approved):
lgtm

Issue #354, opened 2021-04-15
Berlin hard fork seems to fail


florentweiler:
- **OpenEthereum version (>=3.1.0)**: 3.2.2-rc1-stable
- **Operating system**: Debian 10
- **Installation**: binary release
- **Fully synchronized**: no
- **Network**: ethereum
- **Restarted**: yes

Post Berlin hard fork update

Expected behavior:

Blockchain is synchronizing

Actual behavior:

Synchronization is stuck at block 12244294

Additional information:

example of logs : 

`Apr 15 11:25:14 parity-prod-03 parity[18802]: 2021-04-15 11:25:14 UTC Syncing #12244294 0xb2e0…b5a9     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+    0 Qed LI:#12244343   15/25 peers    247 MiB chain  0 bytes queue  RPC:  0 conn,    0 req/s, 1281 µs
Apr 15 11:25:19 parity-prod-03 parity[18802]: 2021-04-15 11:25:19 UTC Syncing #12244294 0xb2e0…b5a9     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+    0 Qed LI:#12244343   14/25 peers    247 MiB chain  0 bytes queue  RPC:  0 conn,    0 req/s, 1281 µs
Apr 15 11:25:48 parity-prod-03 parity[18802]: 2021-04-15 11:25:48 UTC  25/25 peers    249 MiB chain  0 bytes queue  RPC:  0 conn,    2 req/s, 1233 µs
Apr 15 11:26:19 parity-prod-03 parity[18802]: 2021-04-15 11:26:19 UTC  25/25 peers    249 MiB chain  0 bytes queue  RPC:  0 conn,    0 req/s,  837 µs`

Go ethereum status at time of writing : 
synchronized at block 12244383

Etherscan seems to have the same issue

florentweiler:
duplicate with https://github.com/openethereum/openethereum/issues/353
sorry just saw it

PR #356, opened 2021-04-15
EIP-2929: only add params.address to access list

sorpaas:
No matter if this is actually the issue for #354, it looks like a deviation from the specification nontheless. The spec only specify that we add the target address to the access list, but never mentioned `params.sender`.

### DIFF
--- a/crates/ethcore/src/executive.rs
+++ b/crates/ethcore/src/executive.rs
@@ -240,7 +240,6 @@ impl<'a> CallCreateExecutive<'a> {
         if schedule.eip2929 {
             let mut substate = Substate::from_access_list(&params.access_list);
             substate.access_list.insert_address(params.address);
-            substate.access_list.insert_address(params.sender);
             substate
         } else {
             Substate::default()


jochem-brouwer:
image

?

sorpaas:
@jochem-brouwer No. That specifies what happens when you initialize a new transaction, but the lines in this PR is when you enter a substate call/create. They're entirely different.

jochem-brouwer:
Ah - right, in that case you are correct. But this should not change stuff, since `sender` should already be in the warm access lists, as it was previously target of a message call/create or a `*CALL*` / `CREATE*` opcode.

sorpaas:
That's not entirely true. I however indeed hope what you said is correct and this is actually just a no-op deviation.

`params.sender` is the caller of parent contract, not the transaction sender (which is `params.origin`). In the case of `CALLCODE` or `DELEGATECALL`, only `params.code_address` is added to the access list.

jochem-brouwer:
True - sorry, I indeed meant the parent caller here, not `tx.sender` / `ORIGIN`.

jamiemacleanofcoll:
meow

Issue #353, opened 2021-04-15
#12244294 block mainnet issue 


madnadyka:
- **OpenEthereum version**: 3.2.1
- **Operating system**:  Linux
- **Installation**: built from source
- **Fully synchronized**: no
- **Network**: ethereum
- **Restarted**: yes

Node issue after Berlin hard fork, not sync after block 12244294:

2021-04-15 11:15:29 UTC Starting OpenEthereum/v3.2.1-stable-5fdedf0-20210317/x86_64-linux-gnu/rustc1.45.2
2021-04-15 11:15:29 UTC Keys path /root/.local/share/io.parity.ethereum/keys/ethereum
2021-04-15 11:15:29 UTC DB path /root/.local/share/io.parity.ethereum/chains/ethereum/db/906a34e69aec8c0d
2021-04-15 11:15:29 UTC State DB configuration: fast +Trace
2021-04-15 11:15:29 UTC Operating mode: active
2021-04-15 11:15:34 UTC Configured for Ethereum using Ethash engine
2021-04-15 11:15:35 UTC Updated conversion rate to Ξ1 = US$2412.22 (1974075.6 wei/gas)
2021-04-15 11:15:40 UTC Syncing #12244294 0xb2e0…b5a9     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+    0 Qed LI:#12244291    9/50 peers   205 KiB chain 0 bytes queue  RPC:  2 conn,    1 req/s, 357531 µs
2021-04-15 11:15:45 UTC Syncing #12244294 0xb2e0…b5a9     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+    0 Qed LI:#12244291   10/50 peers   6 MiB chain 0 bytes queue  RPC:  2 conn,    2 req/s, 3365 µs
2021-04-15 11:15:45 UTC Accepted a new tcp connection from 185.82.219.84:48124.
2021-04-15 11:15:45 UTC Accepted a new tcp connection from 185.153.197.132:54332.
2021-04-15 11:15:45 UTC eth_accounts is deprecated and will be removed in future versions: Account management is being phased out see #9997 for alternatives.
[...]

PR #360, opened 2021-04-15
EIP-2929: add tx sender and address into the access list

sorpaas:
Possibly related to #353, but again never tested, so let me know if I missed anything.

According to the spec, we should insert the tx address and sender to the access list. This seems to be missing from the code.

### DIFF
--- a/crates/ethcore/src/executive.rs
+++ b/crates/ethcore/src/executive.rs
@@ -1145,6 +1145,7 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
         let mut access_list = AccessList::new(schedule.eip2929);
 
         if schedule.eip2929 {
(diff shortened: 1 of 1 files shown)


adria0:
@sorpaas, the idea is that this is done when the substate is created 
https://github.com/openethereum/openethereum/blob/582bca385fedb1af682e989e5bcc6b3b2cf53028/crates/ethcore/src/executive.rs#L240

adria0 (approved):
Some defensive-programming stuff at this point, but good to check if is redundant or not in the future.

karim-agha (approved):

adria0:
@sorpaas this supersedes `sorpaas:patch-1`?

sorpaas:
@adria0 No. I think rather this goes together with #356

PR #393, opened 2021-05-06
Sunce86/eip 1559

sunce86:
EIP 1559 specification: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

WIP

TODO: 
- [x] Gas target and gas limit handling specification change.
- [ ] ~~Update RPC when 1559 standardization is agreed.~~ Will be a separate PR
- [ ] ~~Tx pool refactor. Make it less generic?~~ Will be a separate PR
- [ ]  ~~Include official json tests.~~ Will be a separate PR
- [ ] ~~TxPermission contract and zero gas price transactions.~~ Will be a separate PR


crates/ethcore/src/machine/impls.rs
```diff
@@ -269,7 +269,7 @@ impl EthereumMachine {
         gas_ceil_target: U256,
     ) {
         header.set_difficulty(parent.difficulty().clone());
-        let gas_limit = parent.gas_limit().clone();
+        let gas_limit = parent.gas_limit() * self.schedule(header.number()).gas_limit_bump;
```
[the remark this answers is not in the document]
rakita: For ethash should we change the gas_limit calculation so that is the same as for other engines. `round_block_gas_limit` seems unnecessary.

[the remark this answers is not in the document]
rakita: Geth uses ceil as target for gas_limit: https://github.com/ethereum/go-ethereum/pull/22896/files#diff-cbeb8282854297a2c8ef02293455cb0258ae75b1d2a1a398b2eb60e59a591ba5R143
and it seems they are doing the same as we are doing for no ethash engine

crates/ethcore/src/miner/pool_client.rs
```diff
@@ -155,9 +155,8 @@ where
     ) -> Result<SignedTransaction, transaction::Error> {
         self.engine
             .verify_transaction_basic(&tx, &self.best_block_header)?;
-        let tx = self
-            .engine
-            .verify_transaction_unordered(tx, &self.best_block_header)?;
```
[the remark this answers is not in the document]
sunce86: I would say yes. 
This part of code is substituted with "let tx = SignedTransaction::new(tx)?;", which is what "verify_transaction_unordered" function was actually doing before this PR.
By doing this substitution I was aiming to avoid base_fee checking for transactions currently being imported into tx pool (since after this PR verify_transaction_unordered does a base_fee checking). 
Maybe there is a more elegant way to achieve the same.

crates/ethcore/src/machine/impls.rs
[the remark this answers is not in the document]
rakita: logic is okey, what I am saying is that bump obscured and hides what is behind by introducing adidtional variable.
```if self.schedule(header.number()).is_london_block() {
    parent.gas_limit* self.params().elestic_multiplier
} else {
    parent.gas_limit
}```
is a lot more straightforward. But in the end it does not matter, it is yours to pick.
please check `round_block_gas_limit`

File: crates/ethcore/src/miner/pool_client.rs (hunk 2 above)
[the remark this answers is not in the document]
Comment (rakita) [reply crates/ethcore/src/miner/pool_client.rs:160]: You are right, i was suspicious but didn't check it.

File: crates/ethcore/res/chainspec/test/london_test.json [hunk 3]
```diff
@@ -0,0 +1,211 @@
+{
+	"name": "Berlin (test)",
```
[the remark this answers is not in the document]
sunce86: Fixed

crates/ethcore/src/executive.rs
```diff
@@ -1472,14 +1491,31 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
         if overflow_1 || overflow_2 {
             return Err(ExecutionError::TransactionMalformed(
                 "U256 Overflow".to_string(),
             ));
         }
 
+        // Up until now, fees_value is calculated for each type of transaction based on their gas prices
+        // Now, if eip1559 is activated, burn the base fee
+        // miner only receives the inclusion fee; note that the base fee is not given to anyone (it is burned)
+        let fees_value = fees_value.saturating_sub(if schedule.eip1559 {
```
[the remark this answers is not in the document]
sunce86: Analyzed together and came to the conclusion that it is not mandatory.

rakita (approved):

crates/ethcore/types/src/header.rs
```diff
@@ -379,14 +406,37 @@ impl Decodable for Header {
...
+            base_fee_per_gas: None,
         };
 
-        for i in 13..r.item_count()? {
-            blockheader.seal.push(r.at(i)?.as_raw().to_vec())
+        if blockheader.number >= eip1559_transition {
+            for i in 13..r.item_count()? - 1 {
+                blockheader.seal.push(r.at(i)?.as_raw().to_vec())
+            }
+            blockheader.base_fee_per_gas = Some(r.val_at(r.item_count()? - 1)?);
```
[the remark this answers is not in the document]
sunce86: Agreed that item_count wont be never zero, wontfix

crates/ethcore/src/machine/impls.rs
```diff
@@ -660,4 +718,68 @@ mod tests {
...
+
+        for i in 0..parent_base_fees.len() {
+            let mut parent_header = Header::default();
+            parent_header.set_base_fee(Some(parent_base_fees[i]));
+            parent_header.set_gas_used(parent_gas_used[i]);
+            parent_header.set_gas_limit(parent_gas_limit[i]);
+
+            let base_fee = machine.calc_base_fee(&parent_header);
+            assert_eq!(expected_base_fee[i], base_fee.unwrap());
+        }
```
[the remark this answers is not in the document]
sunce86: :)

crates/ethcore/src/spec/spec.rs
```diff
@@ -169,6 +173,12 @@ pub struct CommonParams {
     pub transaction_permission_contract_transition: BlockNumber,
     /// Maximum size of transaction's RLP payload
     pub max_transaction_size: usize,
+    /// Base fee max change denominator, EIP1559 related
+    pub base_fee_max_change_denominator: U256,
+    /// Elasticity multiplier, EIP1559 related
+    pub elasticity_multiplier: U256,
+    /// Default value for the block base fee, EIP1559 related
+    pub base_fee_initial_value: U256,
```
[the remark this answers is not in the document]
sunce86: Fixed

crates/ethjson/src/spec/params.rs
```diff
@@ -144,6 +148,12 @@ pub struct Params {
     pub kip4_transition: Option<Uint>,
     /// KIP6 activiation block height.
     pub kip6_transition: Option<Uint>,
+    /// Base fee max change denominator, EIP1559 related
+    pub base_fee_max_change_denominator: Option<Uint>,
+    /// Elasticity multiplier, EIP1559 related
+    pub elasticity_multiplier: Option<Uint>,
+    /// Default value for the block base fee, EIP1559 related
+    pub base_fee_initial_value: Option<Uint>,
```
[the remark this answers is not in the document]
sunce86: Fixed

crates/rpc/src/v1/helpers/dispatch/signing.rs
```diff
@@ -74,6 +75,31 @@ impl super::Accounts for Signer {
                         .collect(),
                 ))
             }
+            Some(TypedTxId::EIP1559Transaction) => {
+                if let Some(max_fee_per_gas) = filled.max_fee_per_gas {
+                    legacy_tx.gas_price = max_fee_per_gas;
+                } else {
+                    return Err(Error::new(ErrorCode::InvalidParams));
+                }
+
```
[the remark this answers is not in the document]
sunce86: Agreed that access list is not mandatory field for eip1559 transactions (for now).

crates/ethcore/src/spec/spec.rs
```diff
@@ -417,6 +444,15 @@ impl From<ethjson::spec::Params> for CommonParams {
             kip6_transition: p
                 .kip6_transition
                 .map_or_else(BlockNumber::max_value, Into::into),
+            eip1559_base_fee_max_change_denominator: p
```
jomuel, line 447: No real complaint, as it is used in only one place and checked for zero there, but an `Option<U256>` would really be preferable here. A divisor with default value zero is awkward.

crates/transaction-pool/src/pool.rs
```diff
@@ -0,0 +1,756 @@
...
+    fn remove_worst(
+        &mut self,
+        transaction: &Transaction<T>,
+        replace: &dyn ShouldReplace<T>,
+    ) -> error::Result<Option<Transaction<T>>, T::Hash> {
+        let to_remove = match self.worst_transactions.iter().next_back() {
+            // No elements to remove? and the pool is still full?
+            None => {
+                warn!("The pool is full but there are no transactions to remove.");
+                return Err(error::Error::TooCheapToEnter(
```
jomuel, line 317: Is `TooCheapToEnter` the right error type here? Same in line 340.

crates/rpc/src/v1/impls/signer.rs
```diff
@@ -124,8 +124,8 @@ impl<D: Dispatcher + 'static> SignerClient<D> {
                             // Altering sender should always reset the nonce.
                             request.nonce = None;
                         }
-                        if let Some(gas_price) = modification.gas_price {
-                            request.gas_price = gas_price;
+                        if let Some(_gas_price) = modification.gas_price {
```
jomuel, line 127:
```suggestion
                        if modification.gas_price.is_some() {
```

Sorry for the bikeshedding :)

jomuel (approved):

crates/transaction-pool/src/pool.rs
sunce86: The original author probably had a wider context on his mind when deciding on the errors. In wider context, function "remove_worst" is called only on the import of the new function. So, if the imported transaction is too cheap, then no transaction should be removed from the pool.

crates/rpc/src/v1/impls/signer.rs
sunce86: Nice one :)
Fixed.

crates/ethcore/src/spec/spec.rs
sunce86: Yes, good point.
Fixed. Please check the fix.

jomuel: Checked it, lgtm

crates/transaction-pool/src/pool.rs
jomuel: Makes sense.

varasev:
Hi guys,

I'm Vadim from xDai team. Responsible for POSDAO consensus.

In OpenEthereum implementation for POSDAO we have a feature which allows to set block gas limit using a smart contract.

The feature was firstly introduced in https://github.com/openethereum/parity-ethereum/pull/10928 and then in the `main` branch for the latest OpenEthereum: https://github.com/openethereum/openethereum/commit/85391f99ac9132c0202afbc619e10a594d2e5112

It is activated by `blockGasLimitContractTransitions` spec option. When it is active, the following rules are ignored:

```
# check if the block changed the gas limit too much
assert block.gas_limit < parent_gas_limit + parent_gas_limit // 1024, 'invalid block: gas limit increased too much'
assert block.gas_limit > parent_gas_limit - parent_gas_limit // 1024, 'invalid block: gas limit decreased too much'
```

For example, see the `gas_limit_override` function and its using in the condition https://github.com/openethereum/openethereum/blob/f9f492638cce143efc5106c1bdf6f4dba5cdff8b/crates/ethcore/src/verification/verification.rs#L479-L491

With the EIP-1559 changes, I think we need to set `eip1559_gas_limit_bump` to `1` if we use the `blockGasLimitContract` for POSDAO. So, if `blockGasLimitContractTransitions` is activated, we have a gas limit value defined by the contract, and on EIP-1559 fork block we shouldn't bump the gas limit by `eip1559_elasticity_multiplier`.

I see the `eip1559_gas_limit_bump` is used [inside the condition `if engine.gas_limit_override(header).is_none()`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/verification/verification.rs#L485) but I also meet it in [`ethcore/src/miner/miner.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/miner/miner.rs#L1452) and [`ethcore/src/machine/impls.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/machine/impls.rs#L272).

Can we be sure here that POSDAO will continue to work with the same block gas limit (defined by the `blockGasLimitContract`) after EIP-1559 activation?

sunce86:
> Hi guys,
> 
> I'm Vadim from xDai team. Responsible for POSDAO consensus.
> 
> In OpenEthereum implementation for POSDAO we have a feature which allows to set block gas limit using a smart contract.
> 
> The feature was firstly introduced in [openethereum/parity-ethereum#10928](https://github.com/openethereum/parity-ethereum/pull/10928) and then in the `main` branch for the latest OpenEthereum: [85391f9](https://github.com/openethereum/openethereum/commit/85391f99ac9132c0202afbc619e10a594d2e5112)
> 
> It is activated by `blockGasLimitContractTransitions` spec option. When it is active, the following rules are ignored:
> 
> ```
> # check if the block changed the gas limit too much
> assert block.gas_limit < parent_gas_limit + parent_gas_limit // 1024, 'invalid block: gas limit increased too much'
> assert block.gas_limit > parent_gas_limit - parent_gas_limit // 1024, 'invalid block: gas limit decreased too much'
> ```
> 
> For example, see the `gas_limit_override` function and its using in the condition
> 
> https://github.com/openethereum/openethereum/blob/f9f492638cce143efc5106c1bdf6f4dba5cdff8b/crates/ethcore/src/verification/verification.rs#L479-L491
> 
> With the EIP-1559 changes, I think we need to set `eip1559_gas_limit_bump` to `1` if we use the `blockGasLimitContract` for POSDAO. So, if `blockGasLimitContractTransitions` is activated, we have a gas limit value defined by the contract, and on EIP-1559 fork block we shouldn't bump the gas limit by `eip1559_elasticity_multiplier`.
> 
> I see the `eip1559_gas_limit_bump` is used [inside the condition `if engine.gas_limit_override(header).is_none()`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/verification/verification.rs#L485) but I also meet it in [`ethcore/src/miner/miner.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/miner/miner.rs#L1452) and [`ethcore/src/machine/impls.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/machine/impls.rs#L272).
> 
> Can we be sure here that POSDAO will continue to work with the same block gas limit (defined by the `blockGasLimitContract`) after EIP-1559 activation?

Hi Vadim,
Thanks for the comment.

[`ethcore/src/machine/impls.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/machine/impls.rs#L272) will be overwritten by [`ethcore/src/engines/authority_round/mod.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/engines/authority_round/mod.rs#L1552) so it won't be an issue.

[`ethcore/src/miner/miner.rs`](https://github.com/openethereum/openethereum/blob/8e995e24776df1d883d3880c12a8c06400ae6ab5/crates/ethcore/src/miner/miner.rs#L1452) might be an issue, so it will be fixed.

crates/ethcore/src/machine/impls.rs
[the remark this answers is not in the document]
sunce86: Gas limit calculation changed to target ceil, can you verify?

crates/ethcore/res/chainspec/test/null_morden.json
```diff
@@ -24,7 +24,7 @@
 		"timestamp": "0x00",
 		"parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
 		"extraData": "0x",
-		"gasLimit": "0x2fefd8"
```
rakita, line 27: Is there a reason why you would change this?

crates/ethcore/src/machine/impls.rs
```diff
@@ -265,59 +262,29 @@ impl EthereumMachine {
...
         assert!(!gas_limit.is_zero(), "Gas limit should be > 0");
 
+        header.set_gas_limit({
+            let bound_divisor = self.params().gas_limit_bound_divisor;
+            if gas_limit < gas_ceil_target {
+                cmp::min(gas_ceil_target, gas_limit + gas_limit / bound_divisor - 1)
+            } else {
+                cmp::max(gas_ceil_target, gas_limit - gas_limit / bound_divisor + 1)
+            }
+        });
```
rakita, line 279: 👍 should be fine

rakita: It just should be communicated that for AuRa/Clique engines `gas_flooor_target` is replaced with `gas_ceil_target`.

rakita: hm, maybe we can even use something like `gas_target = if eip1559 { gas_floor_target} else {gas_ceil_target}` :)

varasev: @rakita How about the `--gas-floor-target` CLI option? Will it be renamed to `--gas-ceil-target`?

varasev:
Hi guys,

For EIP-1559 we also have to take into account the work with TxPermission contract and zero gas price transactions:

`1.` In Nethermind client we introduced a new version of TxPermission (v4) contract which `allowedTxTypes` getter looks like this: https://github.com/poanetwork/posdao-contracts/blob/035d109fd1c810c9f797874608276973aad1780b/contracts/base/TxPermissionV4.sol#L15-L42

So, in the new TxPermission version we replaced the `_gasPrice` parameter with the two params: `_maxFeePerGas` and `_maxPriorityFeePerGas`. And also added the `_gasLimit` parameter (not related to EIP-1559, but will be helpful for our xDai chain).

We agreed that `_maxFeePerGas` and `_maxPriorityFeePerGas` must be equal to `gasPrice` for the legacy transactions.

If we use the previous version of TxPermission (v3), we have the `_gasPrice` parameter there: https://github.com/poanetwork/posdao-contracts/blob/035d109fd1c810c9f797874608276973aad1780b/contracts/base/TxPermissionV3.sol#L36 - we agreed that Ethereum client must pass the `maxFeePerGas` to that parameter if the transaction type is 0x02.

Just in case, the corresponding changes in Nethermind client are here: https://github.com/NethermindEth/nethermind/pull/3043

`2.` For Nethermind we agreed and implemented that the `GASPRICE` opcode (0x3a) should return zero for service transactions (both for legacy and new ones).

`3.` For Nethermind we agreed and implemented that for the new type of transactions (EIP-1559) a service transaction (zero gas price transaction) must have the `maxFeePerGas` field equal to zero, and the `maxPriorityFeePerGas` should be ignored (can be zero or positive). Please note that the service transactions are filtered by the `Certifier` smart contract as always (with its `certified` public getter).

Please add these changes to OpenEthereum as well to make its EIP-1559 behavior the same as in Nethermind.

crates/ethcore/src/machine/impls.rs
rakita: @varasev `--gas-ceil-target` is currently `--gas-cap` option

varasev: Ok, so the above changes with gas target won't affect the current xDai configs, right?

sunce86: > It just should be communicated that for AuRa/Clique engines `gas_flooor_target` is replaced with `gas_ceil_target`.

Yes, do you think that version changelog is not enough, should we announce it directly somewhere?

rakita: @varasev it shouldn't, `gas_limit` is overwritten down the line with `blockGasLimitContract`.
@sunce86 changelog is fine.

sunce86: > hm, maybe we can even use something like `gas_target = if eip1559 { gas_floor_target} else {gas_ceil_target}` :)

Yes, accepted. We should introduce this change for ethash after 1559 activation. Fixed.

> Ok, so the above changes with gas target won't affect the current xDai configs, right?

Updating the configs is not mandatory, but behavior will definitely be different, so aligning the  --gas-floor-target and --gas-cap options with new behavior is desired.

varasev: > @varasev it shouldn't, `gas_limit` is overwritten down the line with `blockGasLimitContract`.

But if we don't use `blockGasLimitContract` (it is not activated on xDai), the behavior will differ from Nethermind?

rakita:
@varasev you are probably more familiar with that part of the code than us, and it seems it is not a big change to add a v4 of the contract, but I think this should be done in a separate pull request. I added you as a member of OpenEthereum, be free to make it.

crates/ethcore/src/machine/impls.rs
rakita: with `gas_target = if eip1559 { gas_floor_target} else {gas_ceil_target}` it will be same even without `blockGasLimitContract`. And in general, setting different `gas_limit` should not cause consensus error.

crates/ethcore/res/chainspec/test/null_morden.json
sunce86: Reverted with the newest commit.

varasev:
> @varasev you are probably more familiar with that part of the code than us, and it seems it is not a big change to add a v4 of the contract, but I think this should be done in a separate pull request. I added you as a member of OpenEthereum, be free to make it.

I agree that this should be done in a separate PR, but I'm not familiar with Rust and most of OE code, so it would be better and faster that someone from OE team would do that

karim-agha (approved):

PR #450, opened 2021-06-25
added effectiveGasPrice to eth_getTransactionReceipt return structure

sunce86:


jomuel (approved):

sunce86:
> this implementation does not conform to the specification
> 
> > the actual value per gas deducted from the senders account. Before EIP-1559, this is equal to the transaction's gas price. After, it is equal to baseFeePerGas + min(maxFeePerGas - baseFeePerGas, maxPriorityFeePerGas). Legacy transactions and EIP-2930 transactions are coerced into the EIP-1559 format by setting both maxFeePerGas and maxPriorityFeePerGas as the transaction's gas price.

Yes, you are right. I've addressed the issue now, PR for changing the spec is initiated.

PR #467, opened 2021-07-02
Implement eth/66

### LINKED ISSUE #465: Implement eth/66
So far we have been relying on eth/65 support by other clients for the rest of OpenEthereum's lifetime. Since geth is planning to drop eth/65 earlier, we will need to implement eth/66.

jomuel:
This is the minimum needed to be eth/66-compatible:

- Request ids from peers are parsed and added to answers
- Own queries to peers get random request ids
- The request ids in the answers by peers are discarded.

Since there are not going to be new OpenEthereum versions with only eth/66, it makes no sense to use request ids internally.

Closes #465


sukri557 (approved):

sunce86 (approved):
lgtm

sukri557 (approved):

PR #484, opened 2021-07-14
feeHistory implementation

sunce86:


crates/rpc/src/v1/impls/eth.rs
```diff
@@ -696,6 +696,208 @@ where
...
+            let is_last = i == last_block + 1;
+
+            if i < pending_block {
+                match get_block_header(i) {
+                    Ok(h) => {
+                        let base_fee = h.base_fee();
+
+                        result.base_fee_per_gas.push(base_fee.unwrap_or_default());
+
+                        if !is_last {
```
jomuel, line 790: Nitpick: I don't think this `if` block is necessary here (Since `i < pending_block <= last_block`, `is_last` cannot be `true`).

crates/rpc/src/v1/impls/eth.rs
```diff
@@ -696,6 +696,208 @@ where
...
+                            result.gas_used_ratio.push(calculate_gas_used_ratio(&h));
+
+                            if reward_percentiles.len() > 0 {
+                                let mut gas_and_reward: Vec<(U256, U256)> = vec![];
+                                if let Some(txs) = get_block_transactions(i) {
+                                    if let Some(receipt) = self.client.block_receipts(&h.hash()) {
+                                        assert!(
```
jomuel, line 797: Do we really need to crash if this fails for some reason?

jomuel (approved):

crates/rpc/src/v1/impls/eth.rs
sunce86: Actually, it is possible for last_block to be far smaller then pending_block (please check how the last_block is calculated, it can be built from some block number deep in history).

crates/rpc/src/v1/impls/eth.rs
sunce86: Valid point. Need some kind of assert, but without crash.

PR #509, opened 2021-08-11
Azure pipeline and docker file to build and store docker image to azure container registry.

tuonghuynh:
Create azure devops pipeline using azure-pipelines.yml. When pipeline run, it will build docker image and store image in azure container registry. Start service using docker run or via azure AKS service.


roninkaizen (changes requested):
the azure/Dockerfile 
contains an error with line 16
rust override 1.52.0
see also
https://github.com/openethereum/openethereum/issues/488

Issue #508, opened 2021-08-10
Low peer counts on v3.3.0-rc6


koen84:
- **OpenEthereum version (>=3.1.0)**: v3.3.0-rc6
- **Operating system**: Ubuntu 20.04 LTS
- **Installation**: binary from releases
- **Fully synchronized**:  yes
- **Network**: ethereum mainnet
- **Restarted**:  yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

I notice very low peer counts across several archive nodes since running v3.3.0-rc6, as in between 0 and 10, while i'm running with these flags `--min-peers 100 --max-peers 250`.  Seems worrisome to get out of sync.

hofry:
It's a real problem, I have a max of 1 or 2 peers
`2021-08-14 22:30:11 UTC   0/25 peers      2 MiB chain  0 bytes queue  RPC:  0 conn,    0 req/s,   16 µs
2021-08-14 22:30:41 UTC   0/25 peers     10 MiB chain  0 bytes queue  RPC:  0 conn,    4 req/s, 1393 µs
2021-08-14 22:30:56 UTC Syncing #13025992 0x845e…06b8     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+   23 Qed LI:#13026016    3/25 peers     72 KiB chain    6 MiB queue  RPC:  0 conn,    1 req/s, 1317 µs
2021-08-14 22:31:16 UTC Syncing #13025999 0x8bb0…76ca     0.70 blk/s  109.1 tx/s   10.2 Mgas/s      0+   16 Qed LI:#13026016    3/25 peers      1 MiB chain    4 MiB queue  RPC:  0 conn,    1 req/s, 1948 µs
2021-08-14 22:31:16 UTC Syncing #13025999 0x8bb0…76ca     0.00 blk/s    0.0 tx/s    0.0 Mgas/s      0+   16 Qed LI:#13026016    3/25 peers      1 MiB chain    4 MiB queue  RPC:  0 conn,    1 req/s, 1948 µs
2021-08-14 22:31:16 UTC Syncing #13025995 0xabc2…50ce     0.60 blk/s   79.8 tx/s    7.6 Mgas/s      0+   20 Qed LI:#13026016    3/25 peers    571 KiB chain    5 MiB queue  RPC:  0 conn,    1 req/s, 1948 µs
[...]

PR #519, opened 2021-09-03
Ignore GetNodeData requests

jomuel:
OpenEthereum nodes keep getting dropped by geth nodes due to invalid NodeData messages. This is likely the cause of issue #508. Just ignoring `GetNodeData` requests instead should be ok for the short remaining lifetime of OpenEthereum, and not result our nodes being dropped by geth.


sunce86 (approved):
lg

LukaszRozmej:
This breaks Nethermind sync on AuRa based networks. We already discussed it, please bring GetNodeDataback for AuRa based networks.

PR #567, opened 2021-11-13
Fix evmbin build

hu55a1n1:
Implements missing trait function `BlockChainClient::is_aura()` for `TestBlockChainClient`.


rimrakhimov:
Hey, wouldn't it be better to implement the function the same way it was implemented in the client (https://github.com/openethereum/openethereum/blob/main/crates/ethcore/src/client/client.rs#L2283)?

Also, could you please change the target branch to _dev_

rimrakhimov:
Closed in #570

PR #578, opened 2021-11-28
Fix for modexp overflow in debug mode

rimrakhimov:


varasev (commented):
Let's add one more unit test here which would check function's behavior if `exp_len` is greater than 1 for the same large `base_len`. Won't `out of range` exception occur in this case for the line `let mut reader = input[(base_len as usize).wrapping_add(96)..].chain(io::repeat(0));`?

rimrakhimov:
May you please clarify, how `exp_len` may impact the index used in the line?

varasev:
> May you please clarify, how `exp_len` may impact the index used in the line?

@rimrakhimov Can `read_exact` get out of range when we read starting from `input[95..]` and input length is 96 and, for example, we are trying to read 10 bytes after 95th byte?

rimrakhimov:
@varasev, oh, I see. Actually, no. The reason is `.chain(io::repeat(0))` part. What it does is simply add infinite number of zeros at the end of the buffer. It can be seen as extending the buffer infinitely. Thus, if there are no more bytes to be read from the `input`, there will be no out of range error, but the reader will just get zeros instead.

varasev:
@rimrakhimov Got it. But let's add the unit test anyway.

rimrakhimov:
Added

varasev:
> Added

Thanks, but this unit test should contain only 96 bytes in the `input` (to try to get out of range).

varasev (approved):

PR #604, opened 2022-01-24
Handling of incoming transactions with maxFeePerGas lower than current baseFee

### LINKED ISSUE #496: Local txs not sending if they are below min-gas-price
For questions please use https://discord.io/openethereum, issues are for bugs and feature requests.

_Before filing a new issue, please **provide the following information**._

- **OpenEthereum version (>=3.1.0)**: v3.3.0rc4
- **Operating system**: Windows
- **Installation**: one-line installer
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

I keep my node at min-gas-price 999000000000 (this is 999 gwei) since I don't care about seeing the mempool. I used to be able to submit txs at any gas price since local txs bypassed the min-gas-price limit. But now my ~30gwei txs do not reach the mempool unless either
1) I lower the min-gas-price to below 30 gwei OR
2) I send at another gasprice higher than min-gas-price

This has only happened since upgrading from v3.2.3 to v3.3.0rc4

FYI I use eth_sendRawTransaction method to send txs

rimrakhimov:
Closes #496.

### Main changes

1. New transactions propagation mechanism was updated. Now, hashes of all new transactions (got whether by RPC call, or from another peer) are added into the special channel. The channel is pooled periodically and corresponding transactions are tried to be propagated to other peers. In addition, the previous propagation mechanism remained, which periodically tries to propagate transactions that are ready to be included into the current block.
2. External transactions verification was updated to allow transactions with effective priority fee lower than user specified ` --min-gas-price` or `--usd-per-tx` options to be added into the pool. For now, the node makes use of maximum specified priority fee instead of effective priority fee as the boundary to reject external transactions.


lordt4ever (approved):
Ura

varasev (approved):

PR #633, opened 2022-03-21
Support for POSDAO contract hardfork

rimrakhimov:
Add rewriteBytecode to AuRa engine parameters and its use at the opening of new blocks.


varasev (approved):
Works fine, but let's keep it unmerged until I check the similar PR for Nethermind client: https://github.com/NethermindEth/nethermind/pull/3889

PR #640, opened 2022-03-25
Additional endpoints

leviska:
Implements https://github.com/openethereum/openethereum/issues/634


bin/oe/rpc.rs
```diff
@@ -35,6 +35,45 @@ pub use parity_rpc::ws::{ws, Server as WsServer};
...
+            return Err(format!("expected host:port|api0;api2, but got: {}", s));
+        }
+
+        let url: Vec<&str> = split[0].split(':').collect();
+        if url.len() != 2 {
+            return Err(format!("expected host:port|api0;api2, but got: {}", s));
+        }
+
+        let apis = if split.len() == 1 {
+            "".into()
```
rimrakhimov, line 64: I believe it should be invalid not to provide at least one API to open

bin/oe/rpc.rs
```diff
@@ -189,44 +244,84 @@ pub fn new_ws<D: rpc_apis::Dependencies>(
...
+                allowed_hosts.clone(),
+                conf.max_connections,
+                rpc::WsExtractor::new(path.clone()),
+                rpc::WsExtractor::new(path.clone()),
+                rpc::WsStats::new(deps.stats.clone()),
+                conf.max_payload,
+            ),
+        )?);
+    }
+
```
rimrakhimov, line 298: Maybe it is possible to eliminate code duplication in lines 219-234;250-263 and 266-297; and 328-332;340-355 and 358-380

bin/oe/rpc.rs
```diff
@@ -152,12 +195,24 @@ pub struct Dependencies<D: rpc_apis::Dependencies> {
     pub stats: Arc<RpcStats>,
 }
 
+fn format_ws_error(url: &str, e: Result<WsServer, rpc::ws::Error>) -> Result<WsServer, String> {
```
rimrakhimov, line 198: IMO, name of the function does not reflect its meaning. I would suggest to change arguments in the following way making it exactly a formatter of ws errors
```suggestion
fn format_ws_error(url: &str, e: rpc::ws::Error) -> String 
```

bin/oe/rpc.rs
```diff
@@ -189,44 +244,84 @@ pub fn new_ws<D: rpc_apis::Dependencies>(
...
+            handler,
+            allowed_origins.clone(),
+            allowed_hosts.clone(),
+            conf.max_connections,
+            rpc::WsExtractor::new(path.clone()),
+            rpc::WsExtractor::new(path.clone()),
+            rpc::WsStats::new(deps.stats.clone()),
+            conf.max_payload,
+        ),
+    )?);
```
rimrakhimov, line 263: If `format_ws_error` arguments to be changed, then it is possible just to use `map_err` for the result of `start_ws`. Also, if code duplication could be eliminated, then we may get rid of `format_ws_error` function completely

rimrakhimov:
@leviska Thank you for the contribution! I liked the implementation in general. 

However found some minor problems and parts which could be improved in my opinion, so added my comments and suggestions above. Besides that, it seems that the client cannot be run with networking interface set to `localhost` which is still highly desirable.

Also, I believe the original issue supposed that where would be only one additional cmd argument `--jsonrpc-additional-endpoints` which serves both `ws` and `http` endpoints. I am totally fine with the way you have divided it into two independent parts. But let ask @varasev also. What do you think about that?

leviska:
> Besides that, it seems that the client cannot be run with networking interface set to localhost which is still highly desirable.

yeah, I found that, but it also doesn't work with `--jsonrpc-interface=localhost` (at least for me?), so I wanted to clarify: do I need to fix this only for additional endpoints, or for usual endpoint too?

varasev:
> Besides that, it seems that the client cannot be run with networking interface set to `localhost` which is still highly desirable.

Maybe it can be run with `local` or `127.0.0.1` for the localhost case?

The docs https://openethereum.github.io/Configuring-OpenEthereum says that we need to use `local` (not `localhost`).

Overall, the `--interface` option, for example, works fine with `local`. We use it in our test setup: https://github.com/poanetwork/posdao-test-setup/blob/4ee7eb69d8ae795b0517c04afac5b39c67c953ce/config/node1.openethereum.toml#L8

> I am totally fine with the way you have divided it into two independent parts. But let ask @varasev also. What do you think about that?

I'm fine with it too.

leviska:
> Maybe it can be run with local or 127.0.0.1 for the localhost case?

Oh yeah, I've missed that function. Now it works for additional endpoints too.

Also fixed all the review notes

bin/oe/rpc.rs
```diff
@@ -189,76 +176,116 @@ pub fn new_ws<D: rpc_apis::Dependencies>(
...
-	}
+
+    let mut endpoints = conf.additional_endpoints;
+    endpoints.push(Endpoint {
+        interface: conf.interface,
+        port: conf.port,
+        apis: conf.apis,
+    });
+
+    let max_connections = conf.max_connections;
```
leviska, line 187: need to unpack the struct here. It works without unpacking in 2021 edition, but doesn't in 2018, see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a4adfd6d0ac4bf09519613391e7a1320

rimrakhimov (approved):
It looks great for me.
@leviska thanks for your work!

The only thing which I may propose is to rename `Endpoint` structure to `RpcEndpoint` to add clarity. Otherwise, I believe we can merge the PR

rimrakhimov:
Also, it may be a great to add some unit tests for function parsing `Endpoint` structure

varasev (commented):
@leviska I guess, for `--jsonrpc-additional-endpoints` and `--ws-additional-endpoints` the value should be in double quotes, right? Example:

["host1:port1|api1;api2,host2:port2|api3"]

At least, I observe this when using `additional_endpoints` option in config `toml` file, e.g.:

```
additional_endpoints = ["all:8641|web3,all:8642|eth"]
```

varasev (approved):
@rimrakhimov Can you approve this?

rimrakhimov:
@varasev Yes, I believe we can merge it

rimrakhimov (approved):

